PR #6, opened 2024-02-01
kernel-devsrc.bbappend: support arm64 kernel 5.19+ on Kirkstone

lucas-akira:
Adapted backport from Langdale.
Add tool files in the kernel sources needed to build kernel 5.19+ on arm64. This should solve the issue where TorizonCore Builder isn't able to build kernel modules on current Verdin AM62 releases of Torizon OS, which have kernel 6.1.

Related-to: TCB-426

### DIFF
--- a/recipes-kernel/linux/kernel-devsrc.bbappend
+++ b/recipes-kernel/linux/kernel-devsrc.bbappend
@@ -17,6 +17,15 @@ do_install:append() {
         chown -R root:root $kerneldir/build/arch/arm64/tools/
         )
     fi
+
+    if [ "${ARCH}" = "arm" ]; then
+        (
+        cd ${S}
+        cp -a --parents arch/arm/tools/gen-sysreg.awk $kerneldir/build/	2>/dev/null || :
+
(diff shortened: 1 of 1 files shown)


EdTheBearded (dismissed):

jsrc27:
On the original commit, I see they also added this for arm

```
cp -a --parents arch/arm/tools/gen-sysreg.awk $kerneldir/build/	2>/dev/null || :
```

I guess we do not need this? I ask since our arm32 modules should also be on a kernel greater than 5.19.

lucas-akira:
Good point, I ended up ignoring that because I thought it referred to powerpc on a quick glance.

I couldn't find `gen-sysreg.awk` in arch/arm/tools/ in [upstream stable 6.1.y](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/arm/tools?h=linux-6.1.y) used in our current arm32 SoMs, but I think we should follow the original commit and put the copy command for arm as well, just to be sure. 

If the file isn't present the copy command is ignored anyway, so that line shouldn't cause problems.

jsrc27 (approved):

PR #22, opened 2024-02-20
Let podman service serve as a docker service

MingliuYan:


recipes-containers/docker/files/docker.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network-online.target docker.socket firewalld.service containerd.service usermount.service
+After=network-online.target docker.socket firewalld.service containerd.service usermount.service time-set.target
```
EdTheBearded, line 4: Hey @MingliuYan, just out of curiosity, why you added time-set.target here?
Just standardazing both services?

MingliuYan: Hey, @EdTheBearded, exactly, I was investigating what services should be run before podman.service and found that, then I thought it'd better docker.service also have it. I knew this is not related to the specific ticket TOR-3370, maybe I shall remove that for this commit.

EdTheBearded: Hey @MingliuYan!
No, no problem, you can leave that, its fine!
Was just curious as to why :grinning:

PR #27, opened 2024-02-25
meta: dont check VARIANT to get container engine

MingliuYan:
VARIANT can be set by end users in image recipes, so it does not always equal to "Docker/Podman", let's avoid checking it to get container engine. Instead, change to check docker/podman binaries on OS, this is safe because podman and docker are conflicting with each other, so they would be present on a same OS.


microhobby (approved):
LGTM

PR #31, opened 2024-02-27
initramfs-framework: add modules to support plymouth on AM62

### LINKED ISSUE #30: Splash screen not working on Verdin AM62
During boot, the splash screen is not displayed on Verdin AM62.

EdTheBearded:
For AM62, plymouth wasn't able to show the splash screen due to the framebuffer not being ready on initramfs.

So we needed to add tidss module (and it's dependencies), pwm and DSI bridges modules to make it work with our DSI to LVDS adapter. With this modification, it also made required that we insert the necessary modules for the DSI to HDMI adapter to work, thats why we also see display_connector and the lontium module.

fixes #30


jsrc27 (approved):
LGTM

PR #33, opened 2024-03-02
Integrate ostree version supporting composefs

### LINKED ISSUE #32: Make required changes to the OS to boot from a composefs image
**Scope**

This ticket applies exclusively to OS images built with ECoT support.

**Goals**

Make changes to the OS so as to cause it to boot with a composefs image as the root mount. This includes:

* Upgrade ostree to a recent version having support for composefs; make sure native builds are possible when composefs support is enabled.
* Make the build system add the required information to the ostree commit in the generated OS image to allow the deployment of a composefs image.
* Make the build system generate the composefs image as part of the initial deployment that goes into the Toradex Easy Installer image.
* Adapt the ramdisk jobs and possibly the root preparation executable (from ostree) in order to mount the composefs image as the root mount the system; consider the bind mounts for the writable directories such as /var, /home and so on.

Depending on the results, we may also decide to deal with the /etc directory on a separate ticket.

Notice that this ticket does not cover ensuring authenticity. We’ll cover this on a separate ticket because that depends on changes in the kernel which in our case will require back-ports.

**Background**

We have previously done some investigation and even a proof-of-concept (PoC) using composefs; in the PoC, the ostree generated deployment (hardlink farm) was still used and composefs was employed to protect only the /usr directory. Unfortunately this approach does not add much protection to the system since the composefs protection can be easily bypassed. One obvious way to do it is by replacing any of the various symlinks existing in the root directory of the deployment. Another way is changing /etc/fstab and overriding the contents of /usr with a bind mount. These two methods would allow an attacker to make persistent changes to the system to execute unsigned code at boot time which would defeat the purpose of secure boot.

The solution to the above problem is what was done by ostree, where the root directory is a mount of a composefs image. Thus, we should try to use inasmuch as possible everything that is currently provided by ostree, possibly adapting things to our use case.

rborn-tx:
This is an initial step for introducing rootfs protection on Torizon OS. This is made of a series of changes to add support to the new implementation of `composefs` as integrated into ostree. It includes:

- Removing remnants from the previous composefs implementation based on a kernel driver.
- Introducing the new version of ostree.
- Adapting the original composefs recipes to the new version.
- Making changes of the initial ramdisk to load appropriate drivers (overlayfs, erofs) and to configure ostree for composefs usage.
- Making changes to the image generation process to produce a composefs image as part of an installer image.

See the individual commit messages for details.

Resolves #32


EdTheBearded (approved):

MingliuYan (approved):
LGTM!

jsrc27 (approved):
Look alright to me as well.

PR #38, opened 2024-03-07
Make /etc transient on the secure boot image

### LINKED ISSUE #37: Make /etc transient on the secure boot image
With the recent integration of ostree version 2024.1 supporting composefs into Torizon OS it became theoretically possible to protect the contents of the /etc directory by making it transient i.e. making all runtime changes volatile so that an attacker would not be able to make long-term changes to the directory aiming, for example, to bypass the protections provided by fs-verity (to be integrated later). However, while testing Torizon OS with a transient /etc, we found that ostree was not able to detect the current/booted deployment which in turn prevented OS updates. Because of this, the transient /etc configuration was left disabled.

Here is the message thrown by ostree in such a situation:

```
$ sudo ostree admin status
error: loading sysroot: Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found
```

Checking commits from version 2024.2, particularly 525a57d21d7c2a3fca011d88b22f461607ab8c25, it seems the issue might have been solved, so that we should be able to enable transient mode now.

Goals

* Update ostree to a newer version having better support to a transient /etc.
* Update related libraries (particularly libcomposefs).
* Ensure that the system can boot with a transient /etc and that OS updates done through the ostree CLI work.
    * NOTE: It might happen that the feature still does not work with a newer version in which case we’ll need to consider how to deal with it.

rborn-tx:
See the issue description for details:

Resolves #37


jsrc27 (approved):
LGTM. Nothing really to add, most of the changes are just updating the OSTree version and refactoring some stuff.

EdTheBearded (approved):

PR #41, opened 2024-03-14
ostree: make OTA updates possible when composefs is employed

### LINKED ISSUE #40: Ensure OTA updates work when composefs is enabled
**Background**

With the latest ostree 2024.4 we found that OTA updates still don't work. Here's the log for one such attempt:

```
aktualizr-torizon[1010]: Current version for ECU ID: ed487cc7aa13c93d960658e6ce725ff2a10f00df35f520173d31ee443c3a5f4f is unknown
aktualizr-torizon[1010]: Error while pulling image: 21 mkdirat: Read-only file system
aktualizr-torizon[1010]: Error of adding remote: unlink(aktualizr-remote.trustedkeys.gpg): Read-only file system
aktualizr-torizon[1010]: Error while pulling image: 21 mkdirat: Read-only file system
aktualizr-torizon[1010]: Download unsuccessful after 3 attempts.
```

**Goal**

Make due amendments to ensure OTA updates work when composefs is enable with a transient /etc.

rborn-tx:
Minor changes that make OTA updates work with secure-boot images having `composefs` enabled.

See the commit message for details.

Resolves #40

### DIFF
--- a/recipes-extended/ostree/ostree-2024.4/0001-deploy-Ensure-boot-directory-is-open-before-accessin.patch
+++ b/recipes-extended/ostree/ostree-2024.4/0001-deploy-Ensure-boot-directory-is-open-before-accessin.patch
@@ -1,0 +1,32 @@
+From 3f8dba21c5913b09e36f76215e77b7d6e7816453 Mon Sep 17 00:00:00 2001
+From: Rogerio Guerra Borin <rogerio.borin@toradex.com>
+Date: Tue, 12 Mar 2024 17:02:58 -0300
+Subject: [PATCH] deploy: Ensure boot directory is open before accessing it
+
+This fixes a bug in the (early) deployment pruning function which before
+tried to access the boot directory without opening it first.
+
+Upstream-Status: Accepted
+
+Signed-off-by: Rogerio Guerra Borin <rogerio.borin@toradex.com>
+---
+ src/libostree/ostree-sysroot-deploy.c | 3 +++
+ 1 file changed, 3 insertions(+)
+
+diff --git a/src/libostree/ostree-sysroot-deploy.c b/src/libostree/ostree-sysroot-deploy.c
+index df1254df..0af76f86 100644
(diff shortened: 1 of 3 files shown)


recipes-extended/ostree/ostree-2024.4/0001-deploy-Ensure-boot-directory-is-open-before-accessin.patch
```diff
@@ -0,0 +1,32 @@
+From 3f8dba21c5913b09e36f76215e77b7d6e7816453 Mon Sep 17 00:00:00 2001
+From: Rogerio Guerra Borin <rogerio.borin@toradex.com>
+Date: Tue, 12 Mar 2024 17:02:58 -0300
+Subject: [PATCH] deploy: Ensure boot directory is open before accessing it
+
+This fixes a bug in the (early) deployment pruning function which before
+tried to access the boot directory without opening it first.
+
+Upstream-Status: Accepted
```
jsrc27, line 9: I see this was accepted upstream as you stated here: https://github.com/ostreedev/ostree/commit/4efb44f35a099c54b9e2895cea9622a97ab6d638

I also see that OSTree just released version 2024.5 that appears to contain this commit: https://github.com/ostreedev/ostree/commits/v2024.5/

Just released ~7hrs ago at the time of me writing this. Would it be better to bump the OSTree version so we get this from upstream rather than having this patch here?

rborn-tx: I'd prefer to do the bump and remove the patch later, otherwise I'd need to retest everything with the new version.

jsrc27: Alright fair enough then. I'll proceed with the merge then.

PR #42, opened 2024-03-22
linux: include erofs scc/cfg from linux-toradex-kmeta.inc

MingliuYan:
linux-torizon.inc is being shared by linux-yocto and linux-toradex recipes, including erofs scc/cfg from there would break linux-yocto build since there is no features/erofs/erofs.scc file in yocto-kernel-cache.

Let's move it to linux-toradex-kmeta.inc.


jsrc27 (approved):
LGTM!

PR #43, opened 2024-03-25
rpcbind: fix a systemd dependency issue

MingliuYan:
rpcbind.service must run after systemd-tmpfiles-setup.service, or else /var/run might not be created yet when rpcbind starts, which leads to a runtime issue.


jsrc27 (approved):
Changes make sense. Though one minor question. @MingliuYan did you run into this issue yourself personally? Or are you just covering the possibility that this might happen?

Just wondering since I don't think I've ever seen this interaction reported before.

PR #44, opened 2024-03-25
rac: fix a invalid SRCREV_FORMAT

MingliuYan:
Fix a invalid SRCREV_FORMAT, SRCREV_FORMAT must be set with the format "foo1_foo2_foo3", the current SRCREV_FORMAT does not have the separator "_".

Set SRCREV_rac, SRCREV_tough, SRCREV_russh accordingly, but disable 'SRCREV_xxx:use-head-next = "${AUTOREV}"' by default, because it does not guarantee work, since the crate dependencies might also need to be updated.

### DIFF
--- a/recipes-sota/rac/rac_git.bb
+++ b/recipes-sota/rac/rac_git.bb
@@ -6,13 +6,26 @@ inherit cargo systemd
 
 # Main source respository
 SRC_URI = " \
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

PR #45, opened 2024-03-25
Draft: support composefs fsverity on qemuarm64

MingliuYan:


MingliuYan:
Hi, Dear maintainers:

This is a Draft MR, to allow some developers try out composefs+fsverity on qemuarm64 machine, please dont merge it.

MingliuYan:
To try out this patch set, on qemuarm64:

in conf/auto.conf, add:
```
DISTROOVERRIDES .= ":cfs-support:cfs-signed"
HOST_SUDO_PASSWORD = "YOUR-SUDO-PASSWORD-ON-BUILD-MACHINE"
```
then:
```
$ bitbake torizon-core-podman
$ runqemu qemuarm64
```

rborn-tx (commented):
This is an interesting implementation. I think we should consider reworking it on top of https://github.com/torizon/meta-toradex-torizon/pull/47 though so as to avoid the constraints I mentioned along this review such as:

- The need for fsverity support on the build machine.
- The need for keeping the fsverity information on the filesystem which brings in all the complication of mounting the wic image and pulling directly into it.

And as a reminder, there are some of the changes here that'll need on the other kernels we use.

What do you think?

rborn-tx:
I'm closing this PR since we're not following this approach of generating verity data at build time, at least in the foreseeable future.

PR #47, opened 2024-03-26
Initial integration of authenticity (fs-verity) capability into the system (disabled atm)

### LINKED ISSUE #46: Integrate the authenticity (fs-verity) capability into the system
Once `fs-verity` is available via `composefs` which requires changes to kernel-space (TOR-3378), we need to do the user-space implementations to integrate it. The final goal is to make the necessary changes so that `fs-verity` is enabled upon rootfs mounting.

The work would include:

* Generating key pairs or allowing users to set them during build.
* Signing the required artifacts upon build.
* Loading the keys at boot-time into the kernel keyring.
* Passing appropriate parameters to the kernel and/or initial ramdisk to ensure verity is enforced on the composefs image.

rborn-tx:
This initial integration covers:

- Enabling commit signing in ostree;
- Generating the keys for signing as part of the build process;
- Adding the composefs image digest to the commit metadata and signing it;
- Saving the verification (public) key into the ramdisk;
- Ensuring fsverity is enabled on the repository upon boot;
- Asking ostree to check the deployment about to be booted with the key.

Please refer to the commit messages for further details.

Note: The commits are intentionally referring to an internal ticket (TOR-3379) because I didn't want GitHub to close #46 yet (until we actually finish the integration, which requires the kernel patches for fsverity).

### DIFF
--- a/classes/image_type_torizon.bbclass
+++ b/classes/image_type_torizon.bbclass
@@ -14,7 +14,11 @@ python() {
         d.setVar('TEZI_CONFIG_FORMAT', torizon_tezi)
 }
 
-TEZI_ROOT_FSOPTS:append:torizon-signed = " -O verity"
+# TODO: Consider always enabling verity.
+#       This would facilitate upgrades from "tdx-signed" to "torizon-signed";
+#       If not done beforehand, the upgrade process would require running
+#       "tune2fs -O verity" on the device.
+TEZI_ROOT_FSOPTS:append:cfs-signed = " -O verity"
 
 python adjust_tezi_artifacts() {
     artifacts = d.getVar('TEZI_ARTIFACTS').replace(d.getVar('KERNEL_IMAGETYPE'), '').replace(d.getVar('KERNEL_DEVICETREE'), '')
--- a/classes/image_type_torizon.bbclass
+++ b/classes/image_type_torizon.bbclass
@@ -117,6 +117,36 @@ EXTRA_OSTREE_COMMIT = " \
 
 IMAGE_CMD:ostreecommit[vardepsexclude] += "EXTRA_OSTREE_COMMIT OSTREE_COMMIT_SUBJECT"
 
+require recipes-extended/ostree/gen-cfs-keys.inc
+
+generate_cfs_keys[lockfiles] += "${DEPLOY_DIR_IMAGE}/cfskeys.lock"
+generate_cfs_keys() {
+    gen_cfs_keys
+}
+
+CFS_OSTREECOMMIT_PREFUNCS_COND ?= " generate_cfs_keys"
+CFS_OSTREECOMMIT_PREFUNCS ?= \
+    "${@d.getVar('CFS_OSTREECOMMIT_PREFUNCS_COND') if 'cfs-signed' in d.getVar('OVERRIDES') else ''}"
+
+CFS_OSTREECOMMIT_DEPENDS_COND ?= "\
+    coreutils-native:do_populate_sysroot \
+    openssl-native:do_populate_sysroot \
+"
+CFS_OSTREECOMMIT_DEPENDS ?= \
+    "${@d.getVar('CFS_OSTREECOMMIT_DEPENDS_COND') if 'cfs-signed' in d.getVar('OVERRIDES') else ''}"
+
+CFS_OSTREECOMMIT_FILE_CHECKSUMS ?= "${@cfs_get_key_file_checksums(d)}"
+
+do_image_ostreecommit[prefuncs] += "${CFS_OSTREECOMMIT_PREFUNCS}"
+do_image_ostreecommit[depends] += "${CFS_OSTREECOMMIT_DEPENDS}"
+do_image_ostreecommit[file-checksums] += "${CFS_OSTREECOMMIT_FILE_CHECKSUMS}"
+
+EXTRA_OSTREE_COMMIT:append:cfs-signed = "\
+    --generate-composefs-metadata \
+    --sign-from-file=${CFS_SIGN_KEYDIR}/${CFS_SIGN_KEYNAME}.sec \
+    --sign-type=ed25519 \
+"
+
 do_image_ostreecommit[postfuncs] += " generate_diff_file"
 generate_diff_file[lockfiles] += "${OSTREE_REPO}/ostree.lock"
 
--- a/classes/image_type_torizon.bbclass
+++ b/classes/image_type_torizon.bbclass
@@ -167,6 +167,16 @@ generate_diff_file () {
     fi
 }
 
+# Enable composefs and fsverity on the deployed ostree repo; setting
+# "ex-integrity.fsverity" to "maybe" rather than "true" is important here
+# because it allows the build to succeed on a host not having fsverity support
+# while causing ostree to enable fsverity upon deployments on the device (which
+# would have the support).
+#
+# TODO: Review this on bumping ostree (this is very likely to change).
+OSTREE_OTA_REPO_CONFIG:append:cfs-support = " ex-integrity.composefs:true"
+OSTREE_OTA_REPO_CONFIG:append:cfs-signed = " ex-integrity.fsverity:maybe"
+
 IMAGE_DATETIME_FILES ??= " \
     ${sysconfdir}/issue \
     ${sysconfdir}/issue.net \
--- a/classes/torizon-signed.bbclass
+++ b/classes/torizon-signed.bbclass
@@ -13,7 +13,24 @@ inherit tdx-signed
 #   the system will be capable of booting from a legacy ostree deployment (based
 #   on hardlinks).
 #
-# - cfs-signed: TBD (not yet implemented)
+# - cfs-signed: Enable the actual ostree commit signing at build time (plus the
+#   addition of the composefs digest into the commit metadata) and the signature
+#   verification at runtime.
 #
-# TODO: Get rid of remaining uses of override "torizon-signed".
-DISTROOVERRIDES:append = ":cfs-support"
+DISTROOVERRIDES:append = ":cfs-support:cfs-signed"
+
+# Variables a user will likely want to set up:
+#
+# - CFS_GENERATE_KEYS: Set to "1" or "0" to enable or disable the composefs key
+#   pair generation at build time.
+#
+# - CFS_SIGN_KEYDIR: Directory where the composefs keys are kept (retrieved from
+#   or generated into when not found (if the key generation is enabled)).
+#
+# - CFS_SIGN_KEYNAME: Base name of the key files; the file holding the secret
+#   key will be named ${CFS_SIGN_KEYNAME}.sec and the one with the public key
+#   will be ${CFS_SIGN_KEYNAME}.pub.
+#
+CFS_GENERATE_KEYS ?= "1"
+CFS_SIGN_KEYDIR ?= "${TOPDIR}/keys/ostree"
+CFS_SIGN_KEYNAME ?= "cfs-dev"
--- a/classes/torizon-signed.bbclass
+++ b/classes/torizon-signed.bbclass
@@ -17,7 +17,8 @@ inherit tdx-signed
 #   addition of the composefs digest into the commit metadata) and the signature
 #   verification at runtime.
 #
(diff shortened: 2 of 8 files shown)


rborn-tx:
@MingliuYan I've added you as reviewer particularly because the work here has a lot of overlap with https://github.com/torizon/meta-toradex-torizon/pull/45. So I thought it'd be important for you to know about it besides having your review. Please take a look at the commit message titled "initramfs-framework: automatically enable fsverity on first boot" where I explained what I believe to be the advantages/disadvantages of the approach being followed with respect to fsverity.

classes/image_type_torizon.bbclass
```diff
@@ -14,7 +14,11 @@ python() {
         d.setVar('TEZI_CONFIG_FORMAT', torizon_tezi)
 }
 
-TEZI_ROOT_FSOPTS:append:torizon-signed = " -O verity"
+# TODO: Consider always enabling verity.
+#       This would facilitate upgrades from "tdx-signed" to "torzion-signed";
+#       If not done beforehand, the upgrade process would require running
+#       "tune2fs -O verity" on the device.
```
jsrc27, line 20: Question about this comment, as it is now, what would happen in such a theoretical update? I imagine it would just fail to boot into the new image and rollback? Or is it something more dramatic than that?

classes/torizon-signed.bbclass
```diff
@@ -13,7 +13,25 @@ inherit tdx-signed
...
+# - CFS_SIGN_KEYDIR: Directory where the composefs keys are kept (retrieved from
+#   or generated into when not found (if the key generation is enabled)).
+#
+# - CFS_SIGN_KEYNAME: Base name of the key files; the file holding the secret
+#   key will be named ${CFS_SIGN_KEYNAME}.sec and the one with the public key
+#   will be ${CFS_SIGN_KEYNAME}.pub.
+#
+CFS_GENERATE_KEYS = "1"
+CFS_SIGN_KEYDIR = "${TOPDIR}/keys/ostree"
+CFS_SIGN_KEYNAME = "cfs-dev"
```
jsrc27, line 37: Are these variables intended to be optionally override by the user? The above comment makes it sound like a user could override these, but a hard assignment is used here which suggests that maybe these are not meant to be changed by the user.

recipes-core/initramfs-framework/files/composefs
```diff
@@ -10,14 +14,105 @@ composefs_error() {
...
+#
+composefs_enable_fsverity_all() {
+	# Enable verity in all repository objects.
+	find "${SYSROOT_DIR}/ostree/repo/objects" -exec fsverity enable '{}' \; 2>/dev/null
+
+	# And also on the composefs file of every deployment.
+	for cfsfile_ in "${SYSROOT_DIR}/ostree/deploy"/*/deploy/*/"${COMPOSEFS_NAME}"; do
+		fsverity enable "${cfsfile_}" 2>/dev/null
+	done
+}
```
jsrc27, line 33: I saw from the commit message that on first boot fsverity can take quite a bit of time on first boot. If the alternative method with `ostree`, as mentioned in the comment, was used instead would this reduce the time? Or not really?

Furthermore, is there any possible optimizations that could be done with regards to this setup time on first boot? Other than adding knowledge of fsverity to the installer and build machine like your commit message stated.

jsrc27 (commented):
For the most part this looks okay to me. Just had some clarifying questions regarding some points in the implementation, but nothing major or blocking.

classes/image_type_torizon.bbclass
rborn-tx: In this case I think the process would be more or less like this:

1) First, upgrade to a version having "optional" support for composefs+verity on the ramdisk.
2) Second, do another update enabling verity; I think the update would fail at this point when pulling the update because libostree wouldn't be able enable verity on the downloaded files; but I'm not totally sure. If it fails, the error is reported to the platform and that's it. Otherwise, the system would finish the download, reboot and probably fail when mounting the root mount which is done by `ostree-prepare-root`. Today, if that happens the system would simply hang in the last step performed by the ramdisk.

Anyways, providing a reliable path for upgrading, which would include allowing rollbacks, is a "to-be-created" ticket in my personal backlog.

classes/torizon-signed.bbclass
rborn-tx: Good catch, I missed the question marks!

recipes-core/initramfs-framework/files/composefs
rborn-tx: The operations done here are exactly the same done by `ostree post-copy` and all the time is taken by the kernel to determine the Merkle tree for each file. Considering that, the only possibility to optimize the total time I can imagine would be trying to parallelize the operation over the various CPU cores. But even this may not help because I see a message saying that the kernel is using the CAAM for determining the hashes so unless the CAAM itself also has independent hashing units, the parallelization may not help.

Thus, if this extra time becomes a problem, I think the best solution would be moving the fsverity enabling to Toradex Easy Installer because then we at least wouldn't need the information to be part of the image (which would require changing the image format, today a simple tarball).

The other option of creating an image with fsverity information would be a big change since it would impact the installer and complicate tooling such as TCB. And it would require the build machine to support fsverity as well, i.e. it'd put constraints on user's machines (I mean those building with Yocto and those using TCB). So, I guess this would be the worst solution.

classes/torizon-signed.bbclass
rborn-tx: Fixed

classes/image_type_torizon.bbclass
jsrc27: Okay then that's good to know. This fsverity stuff introduces a lot of nuances and "gotchas" with regards to update paths and such.

recipes-core/initramfs-framework/files/composefs
jsrc27: I'm not sure if I'm solely capable of deciding whether this initial boot time is a problem or not. In the end it's up to the users. I know some users are quite sensitive regarding boot-time and all, even for one time stuff like this. I suppose they could just do the initial boot in production after flashing so that this initial boot time is not experienced when the device is actually deployed. Though this could also be a complication in terms of production workflows for some users, needing to boot their systems and wait for the initialization to complete on the production floor.

Yeah it's a complex problem with no easy solution I see. In the end it's not a "blocking issue" or anything but certainly something to be aware of going forward.

rborn-tx: Absolutely. Do you think we should ask for PM feedback at this point?

jsrc27: Getting more opinions on the matter wouldn't hurt.

rborn-tx:
@jsrc27 fyi, in the latest commits I implemented a progress bar showing on the console the progress of the fsverity enabling process. This was done to (partially) fulfill the requests I got from internal feedback at Toradex.

jsrc27:
> fyi, in the latest commits I implemented a progress bar showing on the console the progress of the fsverity enabling process. This was done to (partially) fulfill the requests I got from internal feedback at Toradex.

Will this be the only addition from the feedback for now? Or are you planning to add more to this?

Also adding @EdTheBearded to the reviewers so he can take a look at this if he wants.

EdTheBearded:
Everything seems to be ok, the only thing we might want to clear is this boot time increase.
As I understood, this is just on first boot right? Subsequent boots have no increase, or at least not an apparent one?

I really don't see that as a big issue, since it is the first boot only. But a PM opinion on this is valid, and maybe a note on the documentation as well.

rborn-tx:
@EdTheBearded 

> Everything seems to be ok, the only thing we might want to clear is this boot time increase. As I understood, this is just on first boot right? Subsequent boots have no increase, or at least not an apparent one?

Yes, it's just the first boot after installing the OS.

> ... But a PM opinion on this is valid, and maybe a note on the documentation as well.

I've already asked for feedback internally at Toradex and got plenty of it actually. In fact, the progress bar was added as a result of that. An improvement that I proposed was adding feedback via plymouth as well, but that was left for the future. The overall opinion was that the time increase is not a problem as long as we document it well and provide user feedback, so I think we're all aligned.

EdTheBearded (approved):
@rborn-tx awesome!

ldts:
@rborn-tx I am a bit puzzled about how enabling fs-verity on the repo files _after_ the composefs image has been created works: AFAIU, the composefs image needs to already contain the fs-verity digests of the content files and yet, I don't see how the image can be modified during switch root. Could you help me understand what I am missing? where are the digest stored persistently if not in the image?

rborn-tx:
@ldts 

> @rborn-tx so enabling fs-verity during initramfs means that the fs-verity digests wont be taken from the signed composefs image during mount (which I suppose does not open to an attack vector?)

Enabling fs-verity on a file means computing its Merkle tree (a tree of digests/hashes) and storing it somewhere in the filesystem. When we do this operation in the ramdisk we don't change the existing composefs image (part of a ostree deployment) at all; the existing composefs image will contain the "expected" digests of all files and when a file is accessed (via the mounted composefs image) its "measured" digest will be checked against the "expected" one in the composefs image. So, everything should be protected as long as the composefs image itself is protected against modifications.

An important follow up question you might ask is: "How is the composefs image protected?". The answer is: the image is also protected by fs-verity but the expected digest of that file is stored as metadata in a **signed** ostree commit. Note that the expected digest is determined at build time on a developer machine where the commit is also signed.

It's interesting to point out that the code in the ramdisk is also enabling fs-verity on the composefs image, but this is not a security issue because the "measured" digest of this file will be checked by ostree against the "expected" digest stored in a signed ostree commit (if I got it correctly from the ostree code, this check is done when mounting the composefs image).

rborn-tx:
@jsrc27 @EdTheBearded Is there anything missing before we can merge this PR?

ldts:
> Enabling fs-verity on a file means computing its Merkle tree (a tree of digests/hashes) and storing it somewhere in the filesystem. When we do this operation in the ramdisk we don't change the existing composefs image (part of a ostree deployment) at all; the existing composefs image will contain the "expected" digests of all files and when a file is accessed (via the mounted composefs image) its "measured" digest will be checked against the "expected" one in the composefs image. So, everything should be protected as long as the composefs image itself is protected against modifications.

This is the bit I find confusing - since fsverity is not available during build (otherwise we would not be enabling it in initramfs) how are the expected measured digests ever reaching the composefs image?

The way I understood the kernel code, using the ramdisk to enable fs-verity on every file (along with the overlay 'metacopy' attribute) causes the measured digest at enable time to be **copied** to somewhere (not sure where, I guess ext4). Later on, on the first read, the measured digest gests compared against the overlay metacopy.

rborn-tx:
@ldts 

> This is the bit I find confusing - since fsverity is not available during build (otherwise we would not be enabling it in initramfs) how are the expected measured digests ever reaching the composefs image?

The expected digests are determined by libcomposefs and they don't require fs-verity support by the kernel (on the build machine); that library has code to determine the digests and they match what the kernel would compute.

In our case the OS installer image is basically a tarball containing the rootfs which is unpacked by our installer on each device at installation time; in this process any fs-verity information is lost. That's why we re-enable fs-verity on each file; it's not because fs-verity wasn't enabled at build time.

In the end keeping fs-verity info is not strictly needed as long as you can reproduce it and that's what we are doing. By the way, rebuilding that info is a use case foreseen by the ostree guys which motivated the addition of the `ostree admin post-copy` command:

https://github.com/ostreedev/ostree/pull/3094

> The way I understood the kernel code, using the ramdisk to enable fs-verity on every file (along with the overlay 'metacopy' attribute) causes the measured digest at enable time to be copied to somewhere (not sure where, I guess ext4). Later on, on the first read, the measured digest gests compared against the overlay metacopy.

Keep in mind that for composefs to work with fs-verity there were changes made on the kernel side that were implemented by the ostree/composefs guys; I haven't looked at them yet to understand what exactly they do but from the various comments on the ostree Git repo I understand they involve integration between erofs, overlayfs and fs-verity. A kernel without those integrations already allows composefs to work but without the fs-verity part.

ldts:
> The expected digests are determined by libcomposefs and they don't require fs-verity support by the kernel (on the build machine); that library has code to determine the digests and they match what the kernel would compute.

ahh yes you are right, I remember that code now. yes, so it all makes sense. perfect. thanks!

PR #48, opened 2024-03-26
rac: bump SRCREV to latest

### LINKED ISSUE #39: Bump RAC version to fix the bug "does not work due to an empty device key file"
This issue has been fixed on https://github.com/toradex/torizon-rac/pull/11

We need to bump the RAC version in meta-toradex-torizon to get the fix.

jsrc27:
This bump required many of the dependency crates to be updated as well. Most notably the recipe now uses a standard crate for russh and the related dependencies rather than cloning the repo directly.

fixes #39


rborn-tx (approved):
Looks sensible to me.

PR #56, opened 2024-04-22
disable Verdin AM62 Watchdog

EdTheBearded:
During release process we've re-enabled AM62 Watchdog, when we should have waited. The Kernel patch that fixes this issue is not present in the release, so we saw the reboot issues again.

### DIFF
--- a/recipes-core/systemd/systemd-conf_%.bbappend
+++ b/recipes-core/systemd/systemd-conf_%.bbappend
@@ -19,3 +19,7 @@ do_install:append() {
 
 	sed -i "s/@@MACHINE@@/${MACHINE}/g" ${D}${systemd_unitdir}/system.conf.d/10-${BPN}.conf
 }
+
+do_install:append:ti-soc() {
+	sed -i '/^RuntimeWatchdogSec=/d' ${D}${systemd_unitdir}/system.conf.d/10-${BPN}.conf
+}


recipes-core/systemd/systemd-conf_%.bbappend
```diff
@@ -19,3 +19,7 @@ do_install:append() {
 
 	sed -i "s/@@MACHINE@@/${MACHINE}/g" ${D}${systemd_unitdir}/system.conf.d/10-${BPN}.conf
 }
+
+do_install:append:ti-soc() {
+	sed -i '$ d' ${D}${systemd_unitdir}/system.conf.d/10-${BPN}.conf
```
rborn-tx, line 24: That's pretty brittle; how about:

```
sed -i '/RuntimeWatchdogSec=/d' ${D}${systemd_unitdir}/system.conf.d/10-${BPN}.conf
```

EdTheBearded:
> @EdTheBearded Have you identified anything in this set of verity-related commits being reverted that would impact the non-secboot image? If there's anything I'd say it's a bug...

no, nothing like that. but you're right, this shouldn't really affect anything, so we're good.

recipes-core/systemd/systemd-conf_%.bbappend
EdTheBearded: Thanks, I've already changed the command.

rborn-tx (approved):
LGTM

rborn-tx (approved):

EdTheBearded:
@jsrc27 think we could merge?

jsrc27 (approved):
LGTM

PR #59, opened 2024-05-02
Remove go from Torizon master branch

EdTheBearded:
This go recipes was backported some time ago, and now this on the master branch is making us use and older version of go.


jsrc27:
Shouldn't this line also be removed?: https://github.com/torizon/meta-toradex-torizon/blob/master/conf/distro/include/torizon.inc#L94

EdTheBearded:
> Shouldn't this line also be removed?: https://github.com/torizon/meta-toradex-torizon/blob/master/conf/distro/include/torizon.inc#L94

Right, I forgot about that line and the go bbclasses as well. All removed now, thank you!

jsrc27 (approved):

PR #62, opened 2024-05-07
networkmanager: unmanage any uap interface

EdTheBearded:
Since NetworkManager can't play nice with uap interfaces, we're marking them as unmanaged, so this won't affect other connections.

### DIFF
--- a/recipes-connectivity/networkmanager/networkmanager_%.bbappend
+++ b/recipes-connectivity/networkmanager/networkmanager_%.bbappend
@@ -4,6 +4,7 @@ SRC_URI += "\
     file://0001-85-nm-unmanaged.rules-do-not-manage-docker-bridges.patch \
     file://toradex-nmconnection.conf \
     file://network.nmconnection.in \
+    file://99-disable-uap.conf \
 "
 
 # Depend on libedit as it has a more friendly license than readline (GPLv3)
@@ -26,4 +27,8 @@ do_install:append() {
     done
 
     chmod 0600 ${D}${sysconfdir}/NetworkManager/system-connections/network?.nmconnection
+}
+
+do_install:append {
+    install -m 0600 ${WORKDIR}/99-disable-uap.conf ${D}${sysconfdir}/NetworkManager/conf.d/99-disable-uap.conf
 }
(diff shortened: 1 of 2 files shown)


EdTheBearded:
I wanted to hear you opinion on this. Ok, this is something we already have to to whenever we're trying to enable AP on a device, but for this specific connection issue, there might be a possibility of fix.
This is one workaround that was also shared in the issue, so it's not like the issue is a blocker anymore....

I don't really like this and would rather wait to see if we get a proper fix on this. What do you guys think?

recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -27,3 +28,7 @@ do_install:append() {
 
     chmod 0600 ${D}${sysconfdir}/NetworkManager/system-connections/network?.nmconnection
 }
+
+do_install:append:ti-soc() {
```
dolcini, line 32: I think we should do this generally, not just for ti-soc

leograba:
@EdTheBearded, the BSP team accepts the issue as a driver bug.

I would suggest we accept this solution of unmanaging UAP in the short-term, and then revert this once the BSP bug is fixed.

recipes-connectivity/networkmanager/networkmanager_%.bbappend
leograba: Rationale: on BSP 7 this issue will happen for other SoMs that use the mwifiex driver

rborn-tx:
Doing this as a short term solution seems sensible to me, as long as we keep a reminder for undoing it later once the driver problem is solved.

recipes-connectivity/networkmanager/networkmanager_%.bbappend
EdTheBearded: Gotcha. Thanks for the feedback @dolcini, I'll make it generic.

EdTheBearded:
> Doing this as a short term solution seems sensible to me, as long as we keep a reminder for undoing it later once the driver problem is solved.

Maybe we could have a ticket on our backlog to remove this. @leograba?

jsrc27 (approved):
LGTM

PR #64, opened 2024-05-07
image_type_torizon: parse layer information only once

### LINKED ISSUE #61: Parsing of bitbake metadata takes a long time (on the master branch only)
The simple parsing of bitbake metadata is taking a very long time (20-30 minutes) on the master branch. The same doesn't seem to happen on the kirkstone branch.

To clarify: I'm opening this issue here because removing the meta-toradex-torizon from the list of layers in bblayers.conf seems to solve it, so it's likely the issue is in this layer.

rborn-tx:
Obtaining layer information from Git was performed whenever variable EXTRA_OSTREE_COMMIT was expanded which caused metadata parsing to be very slow (20-30 minutes). Since layer information is not supposed to change during build, here we get the layer information only once by using the immediate expansion operator provided by Bitbake. This brings the parsing time to the usual values (1-2 minutes on a good machine).

Resolves #61

### DIFF
--- a/classes/image_type_torizon.bbclass
+++ b/classes/image_type_torizon.bbclass
@@ -94,6 +94,10 @@ def get_layer_revision_information(d):
         e = sys.exc_info()[0]
         bb.warn("Failed to get layers information. Exception: {}".format(e))
 
+# Use immediate expansion here to avoid calling a somewhat costly function whenever
+# EXTRA_OSTREE_COMMIT is expanded.
+OSTREE_LAYER_REVISION_INFO := "${@get_layer_revision_information(d)}"
+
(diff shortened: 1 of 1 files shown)


EdTheBearded (approved):
Wow!

leograba:
Nice! Just for curiosity's sake, do you know why we don't see this huge parsing time on kirkstone?

jsrc27 (approved):
LGTM. Though I am also wondering why we never noticed this slow parsing behavior with Kirkstone.

rborn-tx:
Regarding the question of why the slow parsing doesn't seem to happen on kirkstone: I honestly have no idea :weary:.

PR #67, opened 2024-05-11
draft: dev-tools: drop neofetch in favour of fastfetch

leonheldattoradex:
Neofetch is officially not supported anymore[0], and fastfetch is.

Furthermore, fastfetch has upstream support for Torizon OS [1] and it goes vrooom vrooom super fast.

Here's how it looks

image

[0] https://github.com/dylanaraps/neofetch
[1] https://github.com/fastfetch-cli/fastfetch/blob/dev/src/logo/ascii/torizoncore.txt


leonheldattoradex:
@LinusDierheimer I'm not sure if you're familiar with Yocto, but it's a build system for embedded Linux distros. I proposed to use it here for our binary distro (we used neofetch before), but the "recipe", ie, what actually lets users use fastfetch with Yocto is general enough that everyone using this build system can use it. Let me know if it's interesting for you to have this officially somewhere and I can maintain it.

leonheldattoradex:
Drafted because some 3rd-party dependencies will have to be removed to be compliant with MIT. See https://github.com/fastfetch-cli/fastfetch/issues/533#issuecomment-2097050743.

recipes-devtools/fastfetch/files/LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
jsrc27, line 1: I see that the fastfetch repo has a LICENSE file already. You don't need to include a copy of it here in the layer. Just need to make sure the hash in `LIC_FILES_CHKSUM` matches the hash for the LICENSE file in the source repo.

Unless this was included here for another reason? On quick glance it looks identical to the one from the source repo.

charles2910:
@hiagofranco did package fastfetch for Debian and you guys can take a look at the "makefile" in https://salsa.debian.org/debian/fastfetch/-/blob/debian/latest/debian/rules?ref_type=heads

He also did have to remove [some third party embedded libs](https://salsa.debian.org/debian/fastfetch/-/blob/debian/latest/debian/copyright?ref_type=heads#L7), but I think we'll want to keep libyyjson

leonheldattoradex:
Closed by https://github.com/torizon/meta-toradex-torizon/commit/95eee8f475606c1d5e774ae23c7092dc8ea66cbb

PR #69, opened 2024-05-14
Bring changes introducing composefs from kirkstone into master

rborn-tx:
Besides porting the various commits from kirkstone, resume the work on the rootfs protection by enabling the required kernel features to support composefs (i.e. erofs, fs-verity, etc.).


jsrc27:
Okay not going to lie it's a big diff so I skimmed through the changes. But just to check this is basically equivalent to the work you have done in Kirkstone yes? Plus some adaptations for the newer kernel/versions of other things correct?

jsrc27:
Different topic, I don't think we're building these security options/features with our CI on master are we? Not sure if this was already discussed or not. But it would be helpful to make sure your changes here continue to at least build successfully as master/scarthgap will be more worked on in the near future.

EdTheBearded (approved):
LGTM

jsrc27 (approved):

PR #70, opened 2024-05-20
networkmanager: re-enable NetworkManager on UAP interface

### LINKED ISSUE #63: Re-enable NetworkManager on the UAP interface
The NetworkManager trying to manage the UAP interface on kernels newer than 5.15 led to a bug where the Wi-Fi takes a long time to connect to a network. Due to that, the interface was blacklisted so Wi-Fi client mode works out-of-the-box.

That said, the original issue happened due to the behavior of the mwifiex device driver, therefore when the driver is fixed we must re-enable NetworkManager to manage that interface.

              > Doing this as a short term solution seems sensible to me, as long as we keep a reminder for undoing it later once the driver problem is solved.

Maybe we could have a ticket on our backlog to remove this. @leograba?

_Originally posted by @EdTheBearded in https://github.com/torizon/meta-toradex-torizon/issues/62#issuecomment-2098609781_

EdTheBearded:
Closes #63

### DIFF
--- a/recipes-connectivity/networkmanager/networkmanager/99-disable-uap.conf
+++ b/recipes-connectivity/networkmanager/networkmanager/99-disable-uap.conf
@@ -1,5 +1,0 @@
-[main]
-plugins=keyfile
-
-[keyfile]
-unmanaged-devices=interface-name:uap*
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):
LGTM

PR #77, opened 2024-06-10
auto-provisioning.sh: allow device ID and name to be overridden

rborn-tx:
With this, the auto-provisioning script will source a user defined overrides file named "auto-provisioning-overrides.sh" that is expected to be found in `/etc/sota` or in `/usr/lib/sota`. The sourced file can define functions whose output will determine the device ID (get_device_id) and device name (get_device_name) to be used when registering the device to the platform.


recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
```diff
@@ -157,4 +168,30 @@ function main()
     log "Device successfully provisioned"
 }
 
+# Source the overrides script.
+#
+# Currently the script is expected to expose two functions, namely:
+#
+# - get_device_id: if defined, it should output the "device ID" to be used when
+#   registering the device to the platform.
+# - get_device_name: uf defined, it should output the "device name" to be used
```
jsrc27, line 177: Small typo here "uf defined"

jsrc27 (commented):
Looks alright to me for the most part just a minor comment.

The only other thing I can think of is that in the case `get_device_id` or `get_device_name` from the override script fail, then the entire auto-provisioning fails. But the service will then try again after the retry period is up (default 5mins). If the error from the override script is not something that will get resolved by trying again, then this process will just keep repeating until something/someone intervenes.

I guess this isn't a huge problem, but I guess it's not ideal to keep trying a failed process over and over again. But, then again I can see in some use cases some users might want to keep trying again if the issue is an intermittent one. Maybe we just leave it up to users to modify/override `auto-provisioning.service` if they want to control the specific retry behavior.

Anyways I don't think you need to do anything about this necessarily. I was just thinking this out loud.

recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
rborn-tx: Thanks for the review.

Regarding the error handling, I think all we can do is to document this behavior so a user can add to their overrides script whatever fallback logic they need to ensure a name/ID is always produced even in case the main method fails.

Also, notice that the behavior here is just the same as with the other various errors that cause the provisioning script to fail today, most of which are unlikely to fix themselves. If that was an issue for some user, they could review the service configuration to limit the number of retries. If they did so, a device would require a reboot or a manual restart of the service in order to try again which the user would need to be okay with.

recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
```diff
@@ -13,111 +15,122 @@ PROV_ACCESS_TOKEN=""
 
 PROV_REGISTER_ENDPOINT=${PROV_REGISTER_ENDPOINT:-"https://app.torizon.io/api/accounts/devices"}
 
-function log()
-{
+log() {
```
EstebanSannin, line 18: Hi @rborn-tx I know this is not new, but since I noticed it now I wanted to point it out here.
There is a built-in shell command called “log” that may clash with this function in the future if this command is available in the system.

That said, I don't think we'll ever purposely add this command into our system because we don't need it, but I think it's good to keep in mind that if someday, after updating our shell, we experience strange problems, this could be one of the reasons.

Usually, I use always and underscore before the function name when I'm in doubt, just to be sure that in the future we will avoid all possible clashes.

rborn-tx: @EstebanSannin `log` is not a built-in on Bash and we're explicitly using it as shell here. Anyways, with Bash, even if there was a built-in named `log`, the function with the same name would take precedence. In fact, Bash has commands [builtin](https://www.gnu.org/software/bash/manual/bash.html#index-builtin) and [command](https://www.gnu.org/software/bash/manual/bash.html#index-command) to allow one to explicitly call the built-in or external command respectively in case they have been overridden. 

Thus, I wouldn't expect any issues now or in the future. But, thanks for pointing out your concern :smiley_cat:.

jsrc27 (approved):
LGTM on my end.

EdTheBearded (approved):

PR #87, opened 2024-07-03
setup-environment-internal: fix detection of integration manifests

### LINKED ISSUE #86: Improve Yocto setup script for use-cases where custom manifests are used
Some customers if building Torizon OS with Yocto, add their own manifest to help easily manage their custom layers and such. However, the behavior of our setup script causes an easy to miss effect that can have unexpected results on customer builds. This is an excerpt from our Yocto setup script:

```shell
if [ -L "$MANIFEST_FILE" ]; then
    if [ "$(basename `readlink -f "$MANIFEST_FILE"`)" != "default.xml" ]; then
        INTEGRATION_BUILD=1
    fi
else
    if ! cat $MANIFEST_FILE | grep -q "torizoncore/default.xml"; then
        INTEGRATION_BUILD=1
    fi
fi

if [ "$INTEGRATION_BUILD" = "1" ] && [ "$AUTO_CONF_CREATED" = "0" ]; then
    cat > conf/auto.conf <<EOF
# This is needed when building on integration. With use-head-next you
# always get the newest software (u-boot, kernel, aktualizr-torizon, etc).
# Building on integration without use-head-next your build may fail.
DISTROOVERRIDES .= ":use-head-next"
```

Basically the script checks if the manifest used for this build is default.xml or not. If not it sets use-head-next in the build. This causes the latest commit for various recipes to be fetched instead of the pinned commit. This can be surprising and unexpected behavior for users if they do not know about this interaction. As it would cause their builds to appear to be unstable or not-reproducible.

Community post for reference: https://community.toradex.com/t/imx8mp-u-boot-toradex-recipe-for-torizon-core-docker-not-stable-reproducible/24172

EdTheBearded:
We were presuming that any manifest name other then 'default.xml' was one of our integration manifests, thus activating the 'use-head-next' override. What was also happening was that some clients that created a top level manifest to also sync their own layers, or even Common Torizon (which uses common.xml) were wrongly interpreted as integration manifests.

Now we explicitly check for 'integration.xml' and 'next.xml' as they're our integration layers.

Closes #86

### DIFF
--- a/scripts/setup-environment-internal
+++ b/scripts/setup-environment-internal
@@ -216,11 +216,12 @@ MANIFEST_FILE="${MANIFESTS}"/../manifest.xml
 INTEGRATION_BUILD=0
 
 if [ -L "$MANIFEST_FILE" ]; then
-    if [ "$(basename `readlink -f "$MANIFEST_FILE"`)" != "default.xml" ]; then
+    if [ "$(basename `readlink -f "$MANIFEST_FILE"`)" == "integration.xml" ] ||
+       [ "$(basename `readlink -f "$MANIFEST_FILE"`)" == "next.xml" ]; then
         INTEGRATION_BUILD=1
     fi
 else
-    if ! cat $MANIFEST_FILE | grep -q "torizoncore/default.xml"; then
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

PR #91, opened 2024-07-03
auto-provisioning.service: make it depend on system time sync

### LINKED ISSUE #90: Error in initial call to auto-provisioning causes a 5 min delay before it takes effect
Running the auto-provisioning.sh script results in an initial error.  This triggers the 5 min “RestartSec” from the systemd service script causing a long delay before the devices are available in the platform. It would be nice if this didn’t happen and devices were more quickly adopted into the platform.

With additional logging added we can see what the issue is on initial attempt: 

```
Apr 28 17:42:38 verdin-imx8mp-06849059 auto-provisioning.sh[817]: Starting auto-provisioning script
Apr 28 17:42:38 verdin-imx8mp-06849059 auto-provisioning.sh[817]: Checking provisioning status
Apr 28 17:42:38 verdin-imx8mp-06849059 auto-provisioning.sh[832]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Apr 28 17:42:38 verdin-imx8mp-06849059 auto-provisioning.sh[832]:                                  Dload  Upload   Total   Spent    Left  Speed
Apr 28 17:42:39 verdin-imx8mp-06849059 auto-provisioning.sh[832]: [158B blob data]
Apr 28 17:42:39 verdin-imx8mp-06849059 auto-provisioning.sh[832]: curl: (60) SSL certificate problem: certificate is not yet valid
Apr 28 17:42:39 verdin-imx8mp-06849059 auto-provisioning.sh[832]: More details here: https://curl.se/docs/sslcerts.html
Apr 28 17:42:39 verdin-imx8mp-06849059 auto-provisioning.sh[832]: curl failed to verify the legitimacy of the server and therefore could not
Apr 28 17:42:39 verdin-imx8mp-06849059 auto-provisioning.sh[832]: establish a secure connection to it. To learn more about this situation and
Apr 28 17:42:39 verdin-imx8mp-06849059 auto-provisioning.sh[832]: how to fix it, please visit the web page mentioned above.
```

Seems like the curl SSL certificates are not valid yet when we try to get the access token from the server. Since trying again shortly after boot works, it seems to be some kind of boot initialization/timing issue where we try to use the curl before the certificates are “ready”. If the issue is timing based it may not occur on every type of SOM we have.

This has been easily reproduced on Verdin iMX8M Mini (Drew) and Plus (Jeremias and our customer Fline). See Fline report on https://toradex.slack.com/archives/C06MZRUG70F/p1715337847826789. Here’s also a background on why the timeout is 5 minutes https://toradex.atlassian.net/browse/TOR-2271. Also, we suggest customer edit the timeout https://developer.toradex.com/torizon/torizoncore/production-programming-in-torizon/#check-and-customize-auto-provisioning-process.

EdTheBearded:
Due to the device booting with an older system date, there was a race condition between auto-provisioning and the service that syncs the system time. If auto-provision would run first, the certificates would be invalid, because the current date is prior to theur validity, and the service script would fail.

Closes #90

### DIFF
--- a/recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.service
+++ b/recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Automatically provision the device to the Platform Services
-After=network-online.target
+After=network-online.target systemd-time-wait-sync.service
 Wants=network-online.target
+Requires=systemd-time-wait-sync.service
 ConditionPathExists=/var/sota/auto-provisioning.json
 ConditionPathExists=!/var/sota/import/pkey.pem
 


jsrc27 (approved):
LGTM. I assume this was tested on some modules right?

EdTheBearded:
> LGTM. I assume this was tested on some modules right?

Yup, if you look on our Jira ticket (TOR-2790) I describe how I reproduced and tested the solution. And after built an image with this change, and ran the same tests again and it all worked fine.
All tested on Verdin Plus.

rborn-tx:
LGTM

Looking at the logs on the ticket I noticed the message:

```
Sep 06 15:42:08 verdin-imx8mm-06827382 auto-provisioning.sh[687]: ERROR: HTTP response:                                                                                                                            
Sep 06 15:42:08 verdin-imx8mm-06827382 auto-provisioning.sh[687]: ERROR: Could not read [.access_token] property from the JSON data                                                                                
```

Perhaps we should have a ticket to generally improve the script with regards to providing more useful information in error cases. For example, in this case the HTTP code from the server could have been helpful in pointing out the exact problem...

rborn-tx (approved):

rborn-tx:
Question: would this wait indefinitely when there's no network? Not that think it's a problem if it does.

EdTheBearded:
> Question: would this wait indefinitely when there's no network? Not that think it's a problem if it does.

Hey @rborn-tx, I see you already merged but I ended up checking this anyway.
Yes, the way it is not, the `auto-provisioning.service` will remain inactive if there's not internet connection. And after a connection is established, and the system time is synced, auto-provisioning will run as expected.
Test:
 1. removed ethernet cable from device
 2. changed sw/hw time: `sudo date -s "Tue Sep  6 15:40:51 UTC 2022"; sudo hwclock -w`
 3. changed `auto-provisioning.service` `RestartSec` timeout to 10s
 4. added my credentials to `/var/sota/auto-provisioning.json`
 5. rebooted device
 
 After boot: 
 ```shell
 torizon@verdin-am62-15133443:~$ systemctl status auto-provisioning.service                                                                                                                                         
* auto-provisioning.service - Automatically provision the device to the Platform Services                                                                                                                          
     Loaded: loaded (8;;file://verdin-am62-15133443/usr/lib/systemd/system/auto-provisioning.service/usr/lib/systemd/system/auto-provisioning.service8;;; enabled; vendor preset: enabled)8;;                      
     Active: inactive (dead)
 ```
 
 I waited a few minutes and the service never ran. Then I reconnected the ethernet cable, and after some time the system date was syncronized and the device correctly provisioned:
 ```shell
 torizon@verdin-am62-15133443:~$ journalctl -fu auto-provisioning.service                                                                                                                                           
Jul 08 13:16:30 verdin-am62-15133443 auto-provisioning.sh[971]: Provisioning device with deviceID="verdin-am62-15133443", deviceName="" and downloading credentials
Jul 08 13:16:31 verdin-am62-15133443 auto-provisioning.sh[971]: Updating device credentials
Jul 08 13:16:31 verdin-am62-15133443 auto-provisioning.sh[971]: Restarting aktualizr-torizon
Jul 08 13:16:32 verdin-am62-15133443 auto-provisioning.sh[971]: Enabling and restarting fluent-bit
Jul 08 13:16:32 verdin-am62-15133443 auto-provisioning.sh[1015]: disabled
Jul 08 13:16:32 verdin-am62-15133443 auto-provisioning.sh[971]: Device successfully provisioned
Jul 08 13:16:32 verdin-am62-15133443 auto-provisioning.sh[1019]: Starting auto-provisioning script
Jul 08 13:16:32 verdin-am62-15133443 auto-provisioning.sh[1019]: Checking provisioning status
Jul 08 13:16:32 verdin-am62-15133443 auto-provisioning.sh[1019]: Auto-provisioning configuration file not found [/var/sota/auto-provisioning.json]
Jul 08 13:16:32 verdin-am62-15133443 systemd[1]: Finished Automatically provision the device to the Platform Services.
 ```

rborn-tx:
> Hey @rborn-tx, I see you already merged but I ended up checking this anyway.

@EdTheBearded Thanks for checking. The behavior you describe seems sensible to me.

Other questions that came to mind were:

- What happens if `systemd-timesyncd` is disabled by the user?
- What happens if `systemd-timesyncd` fails to synchronize (e.g. because the device inside some network that does not allow access to that server)?

Anyways, I don't think we'd need to change anything, but depending on the behavior we might need to document something (like "Requirements for auto-provisioning").

EdTheBearded:
> > Hey @rborn-tx, I see you already merged but I ended up checking this anyway.
> 
> @EdTheBearded Thanks for checking. The behavior you describe seems sensible to me.
> 
> Other questions that came to mind were:
> 
>     * What happens if `systemd-timesyncd` is disabled by the user?
> 
>     * What happens if `systemd-timesyncd` fails to synchronize (e.g. because the device inside some network that does not allow access to that server)?
> 
> 
> Anyways, I don't think we'd need to change anything, but depending on the behavior we might need to document something (like "Requirements for auto-provisioning").

If both scenarios happen on the first boot, then it wouldn't matter, since we would have the same thing that was happening on the bug: the certificates would not be valid and the auto-provisioning would continuously fail.
If this happens after a first boot, then I'm not sure. `systemd-timesyncd` is wanted by auto-provisioning, so maybe even if disabled by the user, when systemd reaches auto-provisioning service, it will start.
But if the service starts and is not able to synchronize, I'm not sure what would happen.
Maybe it wouldn't hurt to add a note in the auto-provisioning page stating that it is necessary to time sync in order for this feature to work.

PR #92, opened 2024-07-05
auto-provisioning.sh: add support for hibernated mode

lucas-akira:
The auto-provisioning script will now check in the config file the value of an optional boolean field called `hibernated`. If not present the `false` value is assumed.

The value is then sent to the cloud servers when registering the device. If set to `true` then the device is provisioned in hibernated mode.

This hibernated mode refers to a specific state of a registered device, where it can't receive updates from Torizon Cloud nor send data to it. Devices in this state are not counted towards a Torizon Cloud subscription.

 It is not related to the power state of the device itself.

### DIFF
--- a/recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
+++ b/recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
@@ -12,6 +12,7 @@ PROV_CLIENT_ID=""
 PROV_SECRET=""
 PROV_TOKEN_ENDPOINT=""
 PROV_ACCESS_TOKEN=""
+PROV_HIBERNATED=""
 
 PROV_REGISTER_ENDPOINT=${PROV_REGISTER_ENDPOINT:-"https://app.torizon.io/api/accounts/devices"}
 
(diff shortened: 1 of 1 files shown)


lucas-akira:
I've tested this by building an image with the updated script, provisioned it using TorizonCore Builder and my Torizon Cloud account, appended `"hibernated": true` to `auto-provisioning.json`, then installed the image on an Apalis iMX6.

I also made the same test without appending the hibernation mode string to make sure the script changes don't cause problems with `auto-provisioning.json` config files without the new field.

jsrc27:
So if `hibernated` is a boolean value that is fine. If it does not exist that is fine since the script will force it to a false. What happens if the `hibernated` value exists but it's not a boolean value? Does the provisioning fail?

lucas-akira:
If `hibernated` is sent to the cloud with a non-boolean value, the device will still get provisioned, though without hibernation.

I ended up verifying this behavior when during my tests I accidentally sent the string `"true"` and not the JSON boolean `true` (without parentheses) as the value of `hibernated`.

To avoid this problem I've added a [jq filter](https://jqlang.github.io/jq/manual/#test) called `test` that compares the string and will always return a JSON boolean:
```
[...], "hibernated": $hbrnt | test("true")
```

jsrc27 (approved):

PR #93, opened 2024-07-11
systemd: Add script to recover from rescue or emergency mode

jsrc27:
Several services can trigger rescue or emergency mode on failure. If either mode is triggered after an OTA update the device will require manual intervention. This is not desirable for Torizon OS as updates should be able to recover on their own during failure.

Add a script that will trigger a reboot that will eventually cause a rollback if emergency or rescue mode is triggered while in the middle of an update.

Related-to: TOR-3390


EdTheBearded (approved):
LGTM.
@jsrc27 do we need this on master/scarthgap? If so, could you please cherry-pick this change?

jsrc27:
> do we need this on master/scarthgap? If so, could you please cherry-pick this change?

Probably we do. The rescue/emergency behavior would be the same on these versions as well. I'll need to rebase the patches first though. I know the part of the code-base this patch affects has changed slightly in the latest versions of systemd

jsrc27:
PRs created for master and scarthgap:

* https://github.com/torizon/meta-toradex-torizon/pull/94
* https://github.com/torizon/meta-toradex-torizon/pull/95

PR #100, opened 2024-07-31
Fix genericx86-64 and qemuarm64 builds

EdTheBearded:
Removed a conflicting grub patch that was already present upstream, and fixed a U-Boot build/boot issue, disabling CONFIG_DFU_MTD and changing the `bootcommand` used.


jsrc27 (approved):
LGTM

PR #102, opened 2024-08-02
Fix master build

EdTheBearded:
In order to fix Torizon master build, we needed to refactor a systemd patch to be compliant with v256, and to add Upstream-Status to an ostree patch.


jsrc27 (approved):
LGTM

PR #104, opened 2024-08-07
Revert "u-boot: conditionally disable container generation by U-Boot"

rborn-tx:
This reverts commit 27e034015435e8d4dcc1a0dcd32a6b9d8c71baac. The reason is that we're not going to follow that approach anymore since it has been found that the old container generation (based on the imx-boot recipe) has a very important shortcoming: it is very inefficient in building the container due to extra padding which prevents the SPL from growing even by small amounts (a few kB) which would be a consequence of enabling HAB for example; this issue of extra padding is nicely solved by binman, so we want to build on top of that.


jsrc27 (approved):
LGTM

EdTheBearded (approved):

PR #105, opened 2024-08-13
linux-toradex-kmeta: refactor to be used with BSP's Kernel Cache

EdTheBearded:
BSP has transitioned to used Kernel Cache for their config, and Yocto don't really play nice with multiple Caches, only applying the latest occurence on SRC_URI.

In order for both to coexist, with minimal changes, our Cache top .scc files were renamed, so Yocto won't automatically find it with its KMACHINE and KTYPE variables. Then we manually add this top level for each machine, so that it is applied along with BSP's Cache configs.

Related-to: TOR-3534

### DIFF
--- a/recipes-kernel/linux/linux-toradex-kmeta.inc
+++ b/recipes-kernel/linux/linux-toradex-kmeta.inc
@@ -1,10 +1,13 @@
-SRCREV_meta = "99e38525088a105d6e7d99bd0eb07c9880606d64"
-SRCREV_meta:use-head-next = "${AUTOREV}"
+SRCREV_torizon-meta = "99e38525088a105d6e7d99bd0eb07c9880606d64"
+SRCREV_torizon-meta:use-head-next = "${AUTOREV}"
 
-KMETABRANCH = "kirkstone-6.x.y"
-KMETA = "kernel-meta"
+KMETABRANCH = "scarthgap-7.x.y"
+KMETATORIZON = "torizon-kernel-meta"
 
 KMETAREPOSITORY="github.com/toradex/toradex-kernel-cache.git"
 KMETAPROTOCOL="https"
 
-SRC_URI += "git://${KMETAREPOSITORY};protocol=${KMETAPROTOCOL};type=kmeta;name=meta;branch=${KMETABRANCH};destsuffix=${KMETA}"
+SRC_URI += "git://${KMETAREPOSITORY};protocol=${KMETAPROTOCOL};type=kmeta;name=torizon-meta;branch=${KMETABRANCH};destsuffix=${KMETATORIZON}"
+
+## Compose Torizon top level .scc file and include it to our build.
+KERNEL_FEATURES += "bsp/${MACHINE}-${LINUX_KERNEL_TYPE}-torizon.scc"


EdTheBearded:
This PR is dependent on our internal MR for our Kernel Cache:
https://gitlab.int.toradex.com/rd/torizon-core/toradex-kernel-cache/-/merge_requests/109

recipes-kernel/linux/linux-toradex-kmeta.inc
```diff
@@ -1,10 +1,13 @@
-SRCREV_meta = "99e38525088a105d6e7d99bd0eb07c9880606d64"
-SRCREV_meta:use-head-next = "${AUTOREV}"
+SRCREV_torizon-meta = "99e38525088a105d6e7d99bd0eb07c9880606d64"
```
jsrc27, line 1: Are you not bumping the commit hash yet to the latest on the scarthgap branch?

recipes-kernel/linux/linux-toradex-kmeta.inc
```diff
@@ -1,10 +1,13 @@
...
+SRCREV_torizon-meta:use-head-next = "${AUTOREV}"
 
-KMETABRANCH = "kirkstone-6.x.y"
-KMETA = "kernel-meta"
+KMETABRANCH = "scarthgap-7.x.y"
```
jsrc27, line 4: I merged the internal PR we had. But I don't see the scarthgap branch on our external repo here: https://github.com/toradex/toradex-kernel-cache

Maybe we need to do something to mirror this new branch?

recipes-kernel/linux/linux-toradex-kmeta.inc
EdTheBearded: I could bump, but since we're only building integration for now, we would only be using `AUTOREV`. The hash will only come into play when we do our first release, and we already have a task to bump the hashes there.

recipes-kernel/linux/linux-toradex-kmeta.inc
EdTheBearded: @leograba maybe we have to trigger some manual sync on GitHub?

leograba: I'm a bit overloaded today so it may take a while, but it's at the top of my ToDo list. Will let you know as soon as possible.

leograba: The branch is there now: https://github.com/toradex/toradex-kernel-cache/tree/scarthgap-7.x.y

What was required was to set the new branch as protected on GitLab. Then I manually triggered the mirroring so we didn't need to wait for the next automatic trigger.

What I *didn't* do is make this new branch the default branch either on GitLab or on the GitHub mirror. Please let me know if you'd like me to do it.

EdTheBearded: Thanks @leograba!

>What I didn't do is make this new branch the default branch either on GitLab or on the GitHub mirror. Please let me know if you'd like me to do it.

This is ok. I think we could add this to whenever we do our first release, since at that point we'll switch our default branches.

jsrc27 (approved):

Issue #75, opened 2024-06-07
Master: S = ${WORKDIR} no longer supported


EdTheBearded:
On a [recent commit](https://git.openembedded.org/openembedded-core/commit/?h=master&id=32cba1cc916ad530c5e6630a927e74ca6f06289b) in Styhead's branch, [S = ${WORKDIR} is no longer supported](https://docs.yoctoproject.org/next/migration-guides/migration-5.1.html#s-workdir-no-longer-supported).

We should go over our recipes, and adopt the new recommended pattern, while also checking our custom tasks.
`S = "${WORKDIR}/sources"
UNPACKDIR = "${S}"
`

PR #117, opened 2024-08-23
Add tzn-mqtt recipe

leonheldattoradex:
Clients such as RAC and Aktualizr will can tap into the DBus as listeners and execute commands sent by the cloud over MQTT.

Currently using the dev branch and disabled until full validation and testing is done.

Related-to: TOR-3521


leonheldattoradex:
Same as https://github.com/torizon/meta-toradex-torizon/pull/116 but for master observing https://github.com/torizon/meta-toradex-torizon/issues/75.

recipes-sota/tzn-mqtt/tzn-mqtt_git.bb
```diff
@@ -0,0 +1,434 @@
...
+SRC_URI += " \
+    git://github.com/torizon/tzn-mqtt.git;protocol=https;branch=main \
+    file://tzn-mqtt.service \
+"
+
+SRCREV = "516ce5198a077fffcc2a19d946b3c21814841f01"
+SRCREV:use-head-next = "${AUTOREV}"
+
+S = "${WORKDIR}/sources"
+UNPACKDIR = "${S}"
```
EdTheBearded, line 16: Hey Leon!

This line is not necessary, but is also ok to be here! The only thing is that the way you've done here you're kind of mixing the 2 ways (that I understand) that things work on Styhead+:
1. you add `S = "${WORKDIR}/sources"` and `UNPACKDIR = "${S}"` your recipe. This way, bitbake will perform its `unpac` tasks inside `${S}`. So on our tasks we simply change `${WORKDIR}` for `${S}`.
 
 2. you only add `S = "${WORKDIR}/sources"`. So bitbake will use `UNPACKDIR` default path (`${WORKDIR}/sources-unpack`), and thus we'll need to reference anything from SRC_URI not with `${WORKSPACE}`, but with `${UNPACKDIR}`.
 
 
 Since you've already added `UNPACKDIR = "${S}"` (and I think that was correct, since we're creating this recipe), we can just use `${S}` instead of `${UNPACKDIR}` on line 433.[1]
 
 [1] https://docs.yoctoproject.org/next/migration-guides/migration-5.1.html#s-workdir-no-longer-supported

leonheldattoradex: Good catch, if I understood correctly that makes sense.

I think option 1 is the most clear on intent, I assume you agree from your comment so I did that. Let me know if looks ok.

EdTheBearded (approved):

PR #118, opened 2024-08-26
aktualizr-torizon: fix dependency issue when building SOTA tools

rborn-tx:
When Aktualizr is configured with BUILD_SOTA_TOOLS enabled, it will execute the get-uptane-sign.py script during build which requires a native `python3` interpreter and the `python3-requests` package. Here we amend the Aktualizr recipe to reflect that dependency.

### DIFF
--- a/recipes-sota/aktualizr-torizon/aktualizr-torizon_git.bb
+++ b/recipes-sota/aktualizr-torizon/aktualizr-torizon_git.bb
@@ -24,7 +24,7 @@ PV = "1.0+git${SRCPV}"
 DEPENDS = "boost curl openssl libarchive libsodium sqlite3 asn1c-native ostree"
 RDEPENDS:${PN}:class-target = "aktualizr-hwid lshw bash aktualizr-default-sec aktualizr-polling-interval aktualizr-reboot greenboot"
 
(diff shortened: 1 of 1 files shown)


EdTheBearded (approved):
Oh crap, I forgot about the recipe!
But now I'm wondering how it did not fail on my tests.... Anyways, thanks for spotting that!

@rborn-tx could you please also cherry-pick this into master?

rborn-tx:
> Oh crap, I forgot about the recipe! But now I'm wondering how it did not fail on my tests.... Anyways, thanks for spotting that!

Yeah, the automated builds are not failing either. Anyways, this was detected when building with a CROPS container not having python3-requests installed. I guess the presence of a native installation would mask the issue.

> @rborn-tx could you please also cherry-pick this into master?

Absolutely.

EdTheBearded:
> I guess the presence of a native installation would mask the issue.

Hmm, I was thinking that was the case, since I have it installed natively and on the container I was using.

Just and FYI, I've also bumped our crops, so now you should be able to use `scarthgap-7.x.y`.

jsrc27 (approved):
LGTM

PR #122, opened 2024-08-29
docker-ce: backport patch correcting panic on netlink

EdTheBearded:
Sometimes, especially on devices relying on modems, docker would fail to initialize and create 'docker0' interface. This is a known issue [1], that was solved upstream [2].

Since we're using an older version of docker, I've backported this fix to our layer.

[1] https://github.com/moby/moby/issues/43034
[2] https://github.com/vishvananda/netlink/pull/665

Related-to: TOR-3551

### DIFF
--- a/recipes-containers/docker/docker-ce_git.bbappend
+++ b/recipes-containers/docker/docker-ce_git.bbappend
@@ -3,6 +3,7 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
 SRC_URI:append = " \
     file://0001-dockerd-daemon-use-default-system-config-when-none-i.patch \
     file://0002-cli-config-support-default-system-config.patch \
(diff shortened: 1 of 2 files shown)


recipes-containers/docker/docker-ce_git.bbappend
```diff
@@ -3,6 +3,7 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
 SRC_URI:append = " \
     file://0001-dockerd-daemon-use-default-system-config-when-none-i.patch \
     file://0002-cli-config-support-default-system-config.patch \
+    file://0001-Correct-panic-when-IPv4-lacks-IFA_ADDRESS-address.patch \
```
jsrc27, line 6: Very minor thing, if you could fix the patch numbering so it increments. (i.e. 0003*)

EdTheBearded: Sure, already updated.

jsrc27 (approved):

PR #123, opened 2024-08-29
aktualizr-torizon: fix dependency issue when building SOTA tools

EdTheBearded:
When Aktualizr is configured with BUILD_SOTA_TOOLS enabled, it will execute the get-uptane-sign.py script during build which requires a native python3 interpreter and the python3-requests package. This commit amends the recipe to reflect that dependency.

(cherry picked from commit c24c905)


jsrc27 (approved):
LGTM

PR #125, opened 2024-09-02
networkmanager: unmanage any uap interface

### LINKED ISSUE #108: Re-disable NetworkManager on the UAP interface
**Background**

This is because, even after the BSP fix, customers still see issues with the AP interface when managed by a network manager. Therefore, this time the goal is to change requirements: we won't support NetworkManager (or any other network manager) to manage the AP interface.

Instead, we'll support using lower-level tools directly, `wpa_supplicant` and `hostapd`.

Also, it has been discussed in the NetworkManager project (see https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1529#note_2401201), and unmanaging the interface seems the way to go.

**Goals**

* Un-manage the AP interfaces on Torizon OS (revert https://github.com/torizon/meta-toradex-torizon/issues/63)
* Remove the `NetworkManager` method from the public documentation on https://developer.toradex.com/torizon/application-development/use-cases/networking-connectivity/networking-with-torizoncore/#wi-fi-access-point-mode and keep only the `hostapd` method.
    * Make changes through the article to make other sections (such as the introduction and the hotspot) coherent

EdTheBearded:
Since NetworkManager can't play nice with uap interfaces, we're marking them as unmanaged, so this won't affect other connections.

Related-to: TOR-3545
Closes #108

### DIFF
--- a/recipes-connectivity/networkmanager/networkmanager/99-disable-uap.conf
+++ b/recipes-connectivity/networkmanager/networkmanager/99-disable-uap.conf
@@ -1,0 +1,6 @@
+[main]
+plugins=keyfile
+
+[keyfile]
+unmanaged-devices=interface-name:uap*
+
--- a/recipes-connectivity/networkmanager/networkmanager_%.bbappend
+++ b/recipes-connectivity/networkmanager/networkmanager_%.bbappend
@@ -4,6 +4,7 @@ SRC_URI += "\
     file://0001-85-nm-unmanaged.rules-do-not-manage-docker-bridges.patch \
     file://toradex-nmconnection.conf \
     file://network.nmconnection.in \
+    file://99-disable-uap.conf \
 "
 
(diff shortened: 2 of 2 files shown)


recipes-connectivity/networkmanager/networkmanager/99-disable-uap.conf
```diff
@@ -0,0 +1,5 @@
+[main]
+plugins=keyfile
+
+[keyfile]
+unmanaged-devices=interface-name:uap*
```
rborn-tx, line 5: Nit-pick: missing newline.

EdTheBearded: done! also updated the commits on Scarthgap and Master.

rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #131, opened 2024-09-03
initramfs-framework: remove unnecessary kernel modules from ramfs

EdTheBearded:
After the bump to Kernel 6.6 we got a build error that some kernel modules could not be found. After checking TI Kernel config, either those configs were now builtin or non-existent, so they were removed.

Plymouth (boot splash screen) was tested and is working for both DSI to LVDS and DSI to HDMI.


EdTheBearded:
I'm not cherry-picking this one to master, due to the fact that:
1. I've checked and BSP is still the old 6.1
2. our master build is failing due to other upstream issues, so can't confirm that it's affected

And because that if we get affected, this is now a known issue and is easily fixed.

rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #134, opened 2024-09-09
grub: remove backported patch

EdTheBearded:
This patch was backported, and now that the OE hash was updated to use it we can finally remove.


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #140, opened 2024-09-11
Fix cfs-signed override check

rborn-tx:
Minor fix, though the original code shouldn't ever cause any issues.


EdTheBearded (approved):

jsrc27 (approved):
LGTM

jsrc27:
Does this need to be in master as well?

rborn-tx:
> Does this need to be in master as well?

Good point. Yes, I'll make another PR for that.

PR #146, opened 2024-09-20
avahi: make avahi depend on network-pre target

EdTheBearded:
Since avahi-daemon uses the hostname to create the address, it needs
to be correctly set beforehand.
And since 'set-hostname' service is already ordered to run before
network-pre, we can make sura avahi-daemon only runs aftter that target,
and this way we'll be sure to use the correct hostname.

Related-to: TOR-3562

### DIFF
--- a/recipes-connectivity/avahi/avahi_%.bbappend
+++ b/recipes-connectivity/avahi/avahi_%.bbappend
@@ -1,0 +1,6 @@
+do_install:append() {
+    # On Torizon we use avahi-daemon to setup the hostname. So we need to properly
+    # order it's service to only run after 'set-hostname.service' finishes.
+    sed -i '/^After=/ s/$/ network-pre.target/' ${D}${systemd_system_unitdir}/avahi-daemon.service
+    sed -i '/^Description=/a\Wants=network-pre.target' ${D}${systemd_system_unitdir}/avahi-daemon.service
+}


EdTheBearded:
> @EdTheBearded Shouldn't this dependency be added to the `avahi-daemon` service definition instead (with a `Wants` and `After`)?

Hey @rborn-tx!

Yeah, I can see this solution being done both ways, but let me explain my reasoning for this one.
The way I see this, avahi is properly capturing the available hostname and using it. The thing is that we have a service that sets a different hostname for the device and we want it to be used as it's address.
And with that, the `set-hostname` recipe we use is not the OE one, BSP has overwritten (with a .bb instead of .bbappend) this recipe, which makes it ours, and we already have a .bbappend for this on our layer. So, for me made sense to order those from `set-hostname` perspective, since this is the actual one performing the change.

But if you think it makes more sense to make this change from avahi perspective I'll change it.

rborn-tx:
@EdTheBearded For me adding the dependencies through the `set-hostname` recipe seemed a bit odd because I see the `set-hostname` service as a provider of a functionality that many other services might consume and those other services might be optional. If those services were optional, then I'd like to see each dependency being created only when the optional service is actually used. At the same time, `set-hostname` seems more tied to Torizon OS and thus less likely to ever be optional in any of the Torizon OS variants. Considering this, it seems to me  creating the service dependency from the consumers of the functionality would be better in term of maintenance.

Well, that's my only point but I don't have a strong opinion anyways.

Perhaps we should see what @jsrc27 thinks about it.

jsrc27:
I don't really have a strong opinion either way.

I guess the only other point I can add here is to think about the possible update path. If a user modified one of our base systemd services with an override or something, and then tried to update their OS with OTA/offline updates. Then because of how OSTree works they may not get the updated systemd service from us in our recent OS versions.

Of course I try to advise users not to modify our default systemd services for this exact reason, but it still happens from time to time. The point here being I guess we could take into account whether it is more likely users have modified either the `set-hostname` service or the `avahi-daemon` service.

All that said I don't think it's common customers have modified either really. But just something to think about.

rborn-tx:
Out of curiosity I decided to try to understand why we haven't observed this issue on TOS 6 and I noticed that on TOS 6 we have the following "After" relationships:

```
avahi-daemon.service
   -> systemd-resolved.service
      -> systemd-networkd.service
         -> network-pre.target
            -> set-hostname.service
```

But on Torizon OS 7 the relationship `systemd-resolved.service` -> `systemd-networkd.service` [was removed from systemd](https://github.com/systemd/systemd/commit/b547838000ad640dfd86618202c67680119383a5).

With this, it seems `avahi-daemon` has no direct dependency to the `network` service which explains the issue.

So, one thing to discuss is whether adding the direct dependency to `set-hostname` is the best approach since it's very specific to our use case. A slightly broader solution would be to add the dependency to `network-pre.target` for example.

recipes-connectivity/avahi/avahi_%.bbappend
```diff
@@ -0,0 +1,6 @@
+do_install:append() {
+    # On Torizon we use avahi-daemon to setup the hostname. So we need to properly
+    # order it's service to only run after 'set-hostname.service' finishes.
+    sed -i '/^After=systemd-resolved.service connman.service/ s/$/ network-pre.target/' ${D}${systemd_system_unitdir}/avahi-daemon.service
+    sed -i '/^Description=/a Wants=network-pre.target' ${D}${systemd_system_unitdir}/avahi-daemon.service
```
rborn-tx, line 5: Can we assume GNU `sed` is always used in Yocto? I'm asking because the POSIX way of the `sed` "append" command requires a backslash and a newline as described in https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html:

```
[1addr]a\
text
    Write text to standard output as described previously.
```

Looking at existing OE scripts I see there are some cases where it's assumed the newline is not needed like here: https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-core/initscripts/initscripts_1.0.bb#L110

So, if we follow the same assumption we should at least add the backslash:

```
sed -i '/^Description=/a\Wants=network-pre.target' ${D}${systemd_system_unitdir}/avahi-daemon.service
```

which works with GNU `sed` even if you pass the `--posix` switch. On the other hand, not passing the backslash does not work when `--posix` is passed to GNU `sed` which confirms the fact that the standard requires it.

rborn-tx (commented):
Finished another pass.

recipes-connectivity/avahi/avahi_%.bbappend
```diff
@@ -0,0 +1,6 @@
+do_install:append() {
+    # On Torizon we use avahi-daemon to setup the hostname. So we need to properly
+    # order it's service to only run after 'set-hostname.service' finishes.
+    sed -i '/^After=systemd-resolved.service connman.service/ s/$/ network-pre.target/' ${D}${systemd_system_unitdir}/avahi-daemon.service
```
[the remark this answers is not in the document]
EdTheBearded: No reason, updated it.

recipes-connectivity/avahi/avahi_%.bbappend
EdTheBearded: Ah, nice to know!
I've updated the commit.

rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #147, opened 2024-09-20
avahi: make avahi depend on network-pre target

EdTheBearded:
Since avahi-daemon uses the hostname to create the address, it needs
to be correctly set beforehand.
And since 'set-hostname' service is already ordered to run before
network-pre, we can make sura avahi-daemon only runs aftter that target,
and this way we'll be sure to use the correct hostname.

Related-to: TOR-3562


EdTheBearded:
Updated this commit as well, with changes on avahi side.

rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #148, opened 2024-09-20
set-hostname: fix service dependency on avahi

EdTheBearded:
Since avahi-daemon uses the hostname to create the address, it needs to be correctly set beforehand.
By changing the set-hostname service to run before avahi-daemon service, we eliminate the race condition we have now and guarantee the correct hostname for avahi.

Related-to: TOR-3562


EdTheBearded:
Hey @rborn-tx, thanks to your finding I think we could safely close this PR right?
Since everything is actually ordered in kirkstone, no need to propagate any changes here.

rborn-tx:
> Hey @rborn-tx, thanks to your finding I think we could safely close this PR right? Since everything is actually ordered in kirkstone, no need to propagate any changes here.

Yes, I believe so.

PR #155, opened 2024-09-30
torizon.inc: remove TCLIBCAPPEND variable

EdTheBearded:
In oe-core the TCLIBCAPPEND variable was deprecated, so we need to remove it from our layer.

[1] https://git.openembedded.org/openembedded-core/commit/?id=992ba784c168710328749fd61a0e2869df519dea


jsrc27 (approved):
LGTM

PR #156, opened 2024-10-01
fluent-bit: Setup include files for easy customization.

drewmoseley:
If the users have to modify fluent-bit.conf to add their customizations, that adds a maintenance burden requiring them to merge in changes on every new release.  With this change, user customizations can be stored in seperate include files negaiting that merge step.

### DIFF
--- a/recipes-extended/fluent-bit/files/custom-empty.conf
+++ b/recipes-extended/fluent-bit/files/custom-empty.conf
@@ -1,0 +1,2 @@
+# Empty custom fluent bit config file
+# This exists so that the glob in fluent-bit.conf matches at least one file.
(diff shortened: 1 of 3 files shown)


leograba:
For sure it will require a review on https://developer.toradex.com/torizon/torizon-platform/device-monitoring-in-torizoncore/

@drewmoseley are you doing it?

drewmoseley:
@leograba yes I will add some doc updates for this but probably not until after EW next week.  But I don't think that should hold up merging this PR as this simply adds functionality.  All previous fluent-bit customizations will continue to work.

@EdTheBearded as far as I am concerned this is ready for merge. I've tested it locally. I'll also submit a PR against master but I don't have a working build from that branch.

leograba:
> But I don't think that should hold up merging this PR as this simply adds functionality. All previous fluent-bit customizations will continue to work.

Got it. If it doesn't change the current behavior, it sounds good to even be flexible on the doc update.

EdTheBearded (approved):

rborn-tx (approved):
LGTM

PR #159, opened 2024-10-01
fluent-bit: Setup include files for easy customization.

drewmoseley:
If the users have to modify fluent-bit.conf to add their customizations, that adds a maintenance burden requiring them to merge in changes on every new release.  With this change, user customizations can be stored in seperate include files negaiting that merge step.

### DIFF
--- a/recipes-extended/fluent-bit/files/custom-empty.conf
+++ b/recipes-extended/fluent-bit/files/custom-empty.conf
@@ -1,0 +1,2 @@
+# Empty custom fluent bit config file
+# This exists so that the glob in fluent-bit.conf matches at least one file.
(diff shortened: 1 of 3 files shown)


drewmoseley:
@EdTheBearded here is the master branch version.  Is there an easy way to do a build with this? Do we have a manifest?

EdTheBearded:
Hi @drewmoseley, we do have a manifest for our master builds, but currently it needs some workaround to build, due to some issues on layers we use.

But since this feature was tested on Scarthgap, it is good enough. Since this is a cherry-pick, I would expect this to work just fine as well.

EdTheBearded (approved):

rborn-tx (approved):
LGTM

PR #167, opened 2024-10-25
Add static group, user, and meta-tpm bblayer for fTPM OP-TEE support

tkfu:
fTPM under OP-TEE requires the meta-tpm layer from meta-security to be added to bblayers.conf, and for the tss group and user to exist. Since we use useradd-staticids, we need to add the group and user manually.

### DIFF
--- a/conf/template/bblayers.conf
+++ b/conf/template/bblayers.conf
@@ -42,6 +42,7 @@ BBLAYERS = " \
   ${OEROOT}/layers/meta-toradex-security \
   ${OEROOT}/layers/meta-toradex-distro \
   ${OEROOT}/layers/meta-toradex-bsp-common \
+  ${OEROOT}/layers/meta-security/meta-tpm \
   ${BASELAYERS} \
   ${BSPLAYERS} \
   ${EXTRALAYERS} \
--- a/files/torizon-static-group
+++ b/files/torizon-static-group
@@ -46,6 +46,7 @@ kvm:x:102:
 render:x:103:
 sgx:x:104:
 ptest:x:529:
+tss:x:983:
 systemd-bus-proxy:x:984:
 systemd-timesync:x:985:
 systemd-resolve:x:986:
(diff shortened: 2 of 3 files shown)


tkfu:
Note: I used the highest GID available under 1000, then made the UID equal to that. That was purely for aesthetic reasons, though, and it does leave a gap in the UIDs. We could just as easily do `tss:x:985:983::/var/lib/tpm:/bin/false` if it's preferred to leave no gaps.

EdTheBearded:
Hi @tkfu, is this something you're adding to our regular images or related to our signed ones (meta-toradex-security)?
Just asking to understand if it is a change that could be done on meta-toradex-security instead, but otherwise it seems all good!

tkfu:
As of today, I don't want to put this in our regular images, but I want to make sure that Torizon OS users can make use of meta-toradex-security features without too much trouble.

These are Torizon-specific changes that are required to get the meta-toradex-security fTPM feature working on Torizon OS. They can't really go into meta-toradex-security. The bblayers addition just can't be done in meta-toradex-security at all; the torizon init script just symlinks the local bblayers.conf to meta-toradex-torizon's version. And it'd be quite awkward, and I think an incorrect separation of concerns, to try to modify of Torizon OS's static passwd and groups files in a different layer.

EdTheBearded:
Gotcha, thanks @tkfu!

EdTheBearded (approved):

rborn-tx:
@tkfu With these changes, would the fTPM work without any extra configuration besides enabling it (via variables provided by `meta-toradex-security`)? E.g. would the user need to add anything to `DISTRO_FEATURES`, for example?

jsrc27 (approved):
LGTM as well.

tkfu:
> @tkfu With these changes, would the fTPM work without any extra configuration besides enabling it (via variables provided by `meta-toradex-security`)? E.g. would the user need to add anything to `DISTRO_FEATURES`, for example?

No extra changes required, no! Of course, that's with the caveat that it only works on verdin-imx8mm/verdin-imx8mp as of today.

You do get a warning about the tpm/tpm2 feature not being enabled in DISTRO_FEATURES, but it's not actually a problem because tpm2 is added in MACHINE_FEATURES.

rborn-tx:
> No extra changes required, no! Of course, that's with the caveat that it only works on verdin-imx8mm/verdin-imx8mp as of today.
> 
> You do get a warning about the tpm/tpm2 feature not being enabled in DISTRO_FEATURES, but it's not actually a problem because tpm2 is added in MACHINE_FEATURES.

@tkfu That's great to hear. I asked just to know if we needed some documentation updates. Ok then, changes look good to me.

rborn-tx (approved):

PR #169, opened 2024-10-31
bl_actions.sh: add retry loops to critical operations

rborn-tx:
This is a draft pull request to serve as a possible solution to the bootloader update errors happening on the verdin modules on scarthgap. The actual issue is at the BSP level, so as a first option we want to solve it there. But, if for some reason we need a short term solution, we could use the approach introduced here which basically performs some retries on critical operations (particularly the operation of switching the boot partition).

For details, see the commit message.


rborn-tx:
I'm dropping this because the issue seems to be solved at its root on BSP 7.1.0.

PR #170, opened 2024-11-06
aktualizr-torizon: Exclude aktualizr tmpfiles from systemd cleanup

jsrc27:
Files in /tmp get periodically cleaned-up via systemd-tmpfiles. If this happens to aktualizr related files then, aktualizr will fail to work properly until it gets restarted.

Related-to: TOR-3614


EdTheBearded (approved):

rborn-tx (approved):
This looks sensible to me. Should this be considered as a patch for kirkstone?

jsrc27:
> This looks sensible to me. Should this be considered as a patch for kirkstone?

I was going to ask product/Leo about it as a follow-up after this.

PR #171, opened 2024-11-07
aktualizr-torizon: Exclude aktualizr tmpfiles from systemd cleanup (kirkstone)

jsrc27:
Files in /tmp get periodically cleaned-up via systemd-tmpfiles. If this happens to aktualizr related files then, aktualizr will fail to work properly until it gets restarted.

Related-to: TOR-3614


rborn-tx (approved):
LGTM

EdTheBearded (approved):

PR #175, opened 2024-11-14
cfs-signed: use immediate assignment to DISTROOVERRIDES

rborn-tx:
Make immediate assignment rather using the override syntax to add the cfs-signed and cfs-support overrides to variable DISTROOVERRIDES. This allows one to check that variable in situations where the immediate value is needed, for example in conditional includes/requires.

Minor change just to be consistent with similar changes being done in `meta-toradex-security`:
https://github.com/toradex/meta-toradex-security/pull/83


EdTheBearded (approved):

jsrc27 (approved):
LGTM

PR #176, opened 2024-11-14
initramfs-framework: no longer depend on overlay being a module

rborn-tx:
Make initramfs-module-composefs only recommend rather than strictly depend on initramfs-module-overlay and initramfs-module-erofs. With this, now the corresponding drivers (overlay, erofs) can be compiled as built-in or as modules. Before there was the assumption that those drivers would be always modules but that assumption was broken on scarthgap where the overlay driver became built-in due to changes at the kernel side that in turn lead meta-virtualization to apply certain configuration fragments to the kernel, one of which impacting how the overlay driver was built.

The changes here should only impact images utilizing composefs, which currently means those inheriting class "torizon-signed".


jsrc27 (approved):
LGTM

PR #177, opened 2024-11-18
[Scarthgap] Torizon maintenance

EdTheBearded:
Here’s a summary of the bumps performed in this maintenance:
fastfetch: 2.16.0 → 2.29.0 (release notes: https://github.com/fastfetch-cli/fastfetch/releases)
  \* note: this bump brings us the change from Torizon Core to Torizon OS when running on our OS 
fluentbit: 3.0.7 → 3.1.9 (release notes:https://github.com/fluent/fluent-bit/releases)
  \* note: using meta-openembedded backported recipe, so in the future we can drop this recipe from our layer
greenboot: 0.15.4 → 0.15.7 (release notes: https://github.com/fedora-iot/greenboot/releases)
composefs: 1.0.4 → 1.0.5 (release notes: https://github.com/containers/composefs/releases)
  \* note: using meta-openembedded backported recipe, so in the future we can drop this recipe from our layer


EdTheBearded:
On `composefs` there's no customization to be done, but I've added a .bbappend anyways, in order to remember that this is just a backported recipe that we should drop in the future, and any customization should be done on .bbappend.

But on `fluentbit`, there are some customizations done. Specially on `EXTRA_OECMAKE`. The `PACKAGECONFIG` on the upstream recipe is not complete, and seems to be disabling some default configs of the build. It wasn't straight forward to use only `PACKAGECONFIG` and we still would have to manually modify `EXTRA_OECMAKE` in order to both make it build and configure it to what we had. So in order to simplify things, we just set everything manually on the .bbappend.

leograba:
> Getting closer every day to finally dropping some of these recipes from our meta-layer.

Keep an eye out for stuff we can contribute upstream, guys!

rborn-tx (approved):
LGTM

PR #180, opened 2024-11-21
[Scarthgap] ostree: backport Curl patch

EdTheBearded:
An issue was found during some automated tests, and it was related to OSTree. Looking upstream, there was already an issue opened about this [2], and a fix was recently published [1].
Since this is only available on v2024.9 and were some version behind, we backported this patch.

Related-to: TOR-3649

[1] https://github.com/ostreedev/ostree/pull/3307/commits/4d755a85225ea0a02d4580d088bb8a97138cb040
[2] https://github.com/ostreedev/ostree/issues/3299


recipes-extended/ostree/ostree_%.bbappend
```diff
@@ -8,6 +8,7 @@ SRC_URI:append = " \
     file://0003-ostree-fetcher-curl-set-max-parallel-connections.patch \
     file://0001-mount-Allow-building-when-macro-MOUNT_ATTR_IDMAP-is-.patch \
```
rborn-tx, line 9: nitpick: Shouldn't the new patch be listed here (right after the `0003-ostree-fetcher-...`)?

rborn-tx: Btw, I've just noticed these two patches are already upstream (since v1.0.4 of composefs-tools (now composefs)):

- 0001-mount-Allow-building-when-macro-MOUNT_ATTR_IDMAP-is-.patch
- 0002-mount-Allow-building-when-macro-LOOP_CONFIGURE-is-no.patch

Notice the patches are within the `ostree` recipe but actually apply to its `composefs` submodule.

rborn-tx (changes requested):

recipes-extended/ostree/ostree_%.bbappend
EdTheBearded: Nice, thanks @rborn-tx!

I've added a new commit here removing these two composefs patches, so now the ostree commit has the patches correctly ordered.

rborn-tx (approved):
LGTM

jsrc27 (approved):
LGTM

PR #184, opened 2024-11-25
aktualizr: switch to maintenance branch

rborn-tx:
Switch aktualizr to the new toradex-kirkstone-6.x.y maintenance branch and update the commit to the latest one on that branch. With this, the version of aktualizr used on kirkstone gets the following new commits:

- 3978f3013 Tidy up HttpClient, Provisioner and uptane_test_common
- 6d53c80cd Fetch initial image/root.json straight after provisioning
- dd12f96ff Don't trust initial root.json from a lockbox
- d81b31e9d Increase maximum size of offline-snapshot metadata

Of those, the first one is nice to have while the other three are important to have on kirkstone as part of our maintenance work.


EdTheBearded (approved):

jsrc27 (approved):
LGTM

PR #185, opened 2024-11-26
torizon.inc: use default expected kernel args on verdin-am62 (scarthgap)

rborn-tx:
Update expected kernel arguments to match the ones actually passed to the kernel, which in case of the verdin-am62 now means simply using the default arguments.


EdTheBearded (approved):

EdTheBearded:
@rborn-tx are you cherry-picking this change to master as well?

rborn-tx:
> are you cherry-picking this change to master as well?

@EdTheBearded Sure. I'm also testing it on kirkstone to see if we need the same there.

jsrc27 (approved):
LGTM

PR #191, opened 2024-12-04
Fix for OS rollback not working when kernel image type changes on update

rborn-tx:
This fixes an issue with the rollback feature that happens when the kernel image type changes during an OS update; for example, when one switches from a standard (non secboot) to a secboot image.

See commit messages for further details.


jsrc27 (approved):
LGTM

EdTheBearded (approved):

rborn-tx:
Btw, it seems we don't need this on `kirkstone` but we of course do on the `master`. So, I'll create a follow up PR for that branch.

PR #192, opened 2024-12-05
Fix for OS rollback not working when kernel image type changes on update (master)

rborn-tx:
Same as https://github.com/torizon/meta-toradex-torizon/pull/191, but for the master branch.

Notice I kept the patch number from scarthgap for the moment. Just let me know if you prefer to change it already.


EdTheBearded:
Maybe we could change the numbering already. Even though it's a cherry-pick, I think it's fine to change the file name

jsrc27 (approved):

EdTheBearded (approved):

PR #195, opened 2024-12-16
u-boot-distro-boot: fix do_compile task to generate fuse bank words

EdTheBearded:
The MR that added the fuse bank words [1], also added an issue on do_compile task [2]. Since the arithmetic operator `$((...))` is not supported by bitbake inside a `$(...)` command expression. So 'let' was used to update the for loop index, which is not supported by POSIX shell.

So the fix was to add another task, 'get_fuse_bank_word', made in python and scheduled to run before 'do_compile'. This way, this task processes the BANK_WORD variable, produces the u-boot script string to be replaced later and store it inside 'FUSE_BANK_WORDS'. Then we go into our 'do_compile:append' and simply replace our u-boot script with the newly generated 'FUSE_BANK_WORDS.

Also all bank word variables are now a python tuple, so that we can process them.

[1] https://github.com/torizon/meta-toradex-torizon/pull/190
[2] https://github.com/torizon/meta-toradex-torizon/pull/190/files#diff-4f692e15b504602e32946913e600028ae1ec2e65cfcccfd072283ffd2539c1edR135
/

Related-to: TOR-3663

### DIFF
--- a/recipes-bsp/u-boot/u-boot-distro-boot.bbappend
+++ b/recipes-bsp/u-boot/u-boot-distro-boot.bbappend
@@ -16,24 +16,26 @@ FUSE_NUM = ""
 STATUS_CMD = ""
 CLOSE_CMD = ""
 
-IMX6_COMMON_BANK_WORD = "3 0;\
-3 1;\
-3 2;\
-3 3;\
-3 4;\
-3 5;\
-3 6;\
-3 7;\
+IMX6_COMMON_BANK_WORD = "\
+    (3, 0), \
+    (3, 1), \
+    (3, 2), \
+    (3, 3), \
+    (3, 4), \
+    (3, 5), \
+    (3, 6), \
+    (3, 7) \
 "
 
-IMX7_IMX8M_COMMON_BANK_WORD = "6 0;\
-6 1;\
-6 2;\
-6 3;\
-7 0;\
-7 1;\
-7 2;\
-7 3;\
+IMX7_IMX8M_COMMON_BANK_WORD = "\
+    (6, 0), \
+    (6, 1), \
+    (6, 2), \
+    (6, 3), \
+    (7, 0), \
+    (7, 1), \
+    (7, 2), \
+    (7, 3) \
 "
 
 FUSE_NUM_8 = "1 2 3 4 5 6 7 8"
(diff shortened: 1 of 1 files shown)


EdTheBearded:
This is how the uEnv.txt file is being generated:
```
kernel_image_type=zImage                                                                                 
overlays_file="overlays.txt"                                                                             
otaroot=1                                                                                                
fitconf_fdt_overlays=""
fuse_bank_word_1=3 0;
fuse_bank_word_2=3 1;
fuse_bank_word_3=3 2;
fuse_bank_word_4=3 3;
fuse_bank_word_5=3 4;
fuse_bank_word_6=3 5;
fuse_bank_word_7=3 6;
fuse_bank_word_8=3 7;

fuse_num=1 2 3 4 5 6 7 8
...
```

Tested on Apalis iMX6.

rborn-tx:
It's incredible that bitbake does not allow the POSIX `$((...))`:

https://lists.yoctoproject.org/g/poky/topic/causes_parsing_error/84920289

rborn-tx (commented):
Nice solution. I'm suggesting some changes though.

recipes-bsp/u-boot/u-boot-distro-boot.bbappend
```diff
@@ -129,19 +133,23 @@ UENV_VARS_TO_DEL_EXTRA ?= "kernel_image \
...
-    bank_word=$(echo "$bank_word" | sed -n -e '1h;2,$H;${g;s/\n/\\n/gp}')
+python get_fuse_bank_word() {
+    bank_word_in=list(eval(d.getVar("BANK_WORD")))
+    bank_word=""
+    for i, bw in enumerate(bank_word_in, start=1):
+        bank_word+="fuse_bank_word_{}={} {};\\n".format(i, bw[0], bw[1])
+    d.setVar("FUSE_BANK_WORDS", bank_word)
+}
+do_compile[prefuncs] += "get_fuse_bank_word"
+get_fuse_bank_word[doc] = "Process BANK_WORD variable to embedd into u-boot script."
```
rborn-tx, line 144: Some suggestions:

- Avoid changing the general format of the input.
- Avoid using a prefunc; at least I don't see the need for it if we simply set another variable from `BANK_WORD`.
- Since the value in `FUSE_BANK_WORD` is very specific (it's the string to the added to uEnv.txt already formatted to be used by `sed`), I'd use a name indicating it's usage is somehow internal e.g. `BANK_WORD_INTERNAL`.

Something like this:

```
def get_bank_word_internal(d):
    bank_word_in = d.getVar("BANK_WORD").split(";")
    bank_word = ""
    for i, bw in enumerate(bank_word_in, start=1):
        bank_word += "fuse_bank_word_{}={};\\n".format(i, bw)
    return bank_word

BANK_WORD_INTERNAL = "${@get_bank_word_internal(d)}"

# then use BANK_WORD_INTERNAL in do_compile()
```

We'd need to remove the last `;` from `BANK_WORD` with this implementation though.

What do you think?

EdTheBearded: I think itens 2 and 3 do make sense, and I'll submit the changes.
About 1, I changed the format of the variables for 2 reasons:
 1. Since BANK_WORD is a string, whenever we parse we would get all the words in the string. For instance, with the split on ';' we would get: '3', ' ', '0', '3', ' ', '1'...
 2. Since this is an arbitrary variable that we create and consume in the same file.
So because of this, I decided to already initialize those as a tuple, since it would facilitate the processing.

rborn-tx: I'm not sure I understand what you mean here. The result of `"3 0;3 1;...;3 7".split(";")` is `['3 0', '3 1', ..., '3 7']` which seems to be exactly what we need, no?

EdTheBearded: With `.split(";")`, it was adding a last empty line.
```
WARNING: u-boot-distro-boot-1.0-r0 do_compile: internal fuse_bank_word_1=3 0;
fuse_bank_word_2= 3 1;                                                                                   
fuse_bank_word_3= 3 2;                                                                                   
fuse_bank_word_4= 3 3;                                                                                   
fuse_bank_word_5= 3 4;                                                                                   
fuse_bank_word_6= 3 5;                                                                                   
fuse_bank_word_7= 3 6;                                                                                   
fuse_bank_word_8= 3 7;
fuse_bank_word_9=;
```

So, since how we setup these variables are not that important, I'll keep the tuple format. Just so we can fix this issue and go back to having our nightlies.

rborn-tx: np, approving it

rborn-tx (approved):

jsrc27 (approved):

PR #196, opened 2024-12-18
u-boot-distro-boot: fix 'get_bank_word_internal' with empty strings

EdTheBearded:
For unsupported machines, BANK_WORD is an empty string and since this was not being handled on this function we had an exception in this function that caused the build to fail.

Now if BANK_WORD is empty, we return early with "" as well.

Related-to: TOR-3664


jsrc27 (approved):
LGTM

jsrc27:
> Also, was the original change also merged to master branch?

No, it was not merged there yet, probably should though.

EdTheBearded:
Added a new commit removing the `bbwarn` I forgot.

EdTheBearded:
@jsrc27, @rborn-tx any comments here or we can merge this?
This is failing our builds for TI and qemu machines.

rborn-tx (approved):
LGTM

PR #197, opened 2024-12-22
Add support for upgrading from BCoT to ECoT

rborn-tx:
The main difference between BCoT (basic chain of trust) and ECoT (extended chain of trust) is the addition of the root filesystem protections to ensure integrity and authenticity, both of which achieved by the use of composefs.

In this PR we add a the capability of upgrading a legacy OSTree installation (based on hard links) to one backed by composefs. Overall, the upgrade method introduced here involves:

- Implementation inside the initial ramdisk to enable fsverity on the filesystem and on the individual files in the OSTree repository.
- A systemd service responsible for changing the OSTree repository configuration (which is itself kept outside of OSTree) to enable both composefs and fsverity on the OSTree/libostree side.

For further details, please refer to the commit messages.

**NOTE**: The actual upgrade will require very specific steps which will be documented later.

### DIFF
--- a/recipes-core/initramfs-framework/files/composefs
+++ b/recipes-core/initramfs-framework/files/composefs
@@ -2,6 +2,12 @@
 # Copyright (C) 2023 Toradex AG.
 # Licensed on MIT
 
+# Enable additions to allow upgrades from legacy to composefs-based storage (OSTree).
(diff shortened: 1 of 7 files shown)


EdTheBearded (approved):

EdTheBearded:
@rborn-tx is this going to be necessary on master as well?
Could you please cherry pick this PR?

PR #205, opened 2025-01-07
[Kirkstone] base-passwd: change base-passwd bbappend to match any version

EdTheBearded:
Our bbappend recipe was appending only to a specific version of base-passwd recipe, and since this recipe was bumped on OE upstream, we were missing our group patched and having a build failure.

Backported from Scarthgap the 2 patches that add the necessary groups and also corrects some GIDs to the expected values on Torizon.

Also removed unused patches input.patch and kvm.patch.

Related-to: TOR-3675


jsrc27 (approved):
LGTM

PR #213, opened 2025-01-25
Attempt to solve "Kernel modules not working correctly on preempt-rt images"

rborn-tx:
This is an attempt to solve an issue with the PREEMPT-RT builds where the build of the kernel (along with device-trees and overlays) does not match the build of the kernel modules. Specifically, what does not match is the string responsible for defining the kernel ABI version which in turn also defines the name of the directory containing the kernel and kernel modules.

To try and solve the issue, here we modify the do_patch task to make it forcefully set the commit date of the last commit which is supposed to be the commit applying the RT patches. Normally applying those patches produces different hashes every time because the said patches do not have date information in then leading the kgit-s2q tool to use the current date on the commit produced by the patching operation.

Related-to: TOR-3656

To clarify, here I say "attempt to solve" because I wasn't able to reproduce the problem with my builds. So this work must still be validated with the actual production CI/CD environment.

### DIFF
--- a/recipes-kernel/linux/linux-torizon.inc
+++ b/recipes-kernel/linux/linux-torizon.inc
@@ -10,6 +10,32 @@ SRC_URI:append:cfs-signed = "\
 
 KBUILD_BUILD_VERSION .= "-Torizon"
 
+do_patch:append:preempt-rt() {
+	# When the RT patches are applied via "git apply" by the kgit-s2q tool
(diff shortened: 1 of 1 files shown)


EdTheBearded (approved):
:scream: :feelsgood:

jsrc27 (approved):

rborn-tx:
With the approvals confirmed, I'll go ahead and merge this.

PR #217, opened 2025-01-30
auto-provisioning: improve behavior with power cuts

jsrc27:
Some users when auto-provisioning, power down the system once they see the auto-provisioning service exit successfully. However, in some cases this can cause the files created from the provisioning process to not be fully written to disk, resulting in 0-byte files upon next boot.

Therefore, add an explicit sync after the files are written to improve the behavior of the auto-provisioning in these cases.

Related-to: TOR-3743

### DIFF
--- a/recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
+++ b/recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
@@ -136,16 +136,24 @@ register_device() {
 }
 
 write_credentials() {
+    local temp_dir
+
     log "Updating device credentials"
 
-    rm -Rf $SOTA_CRED_DIR && mkdir -p $SOTA_CRED_DIR
-    if ! unzip device.zip -d $SOTA_CRED_DIR >/dev/null; then
+    temp_dir="${SOTA_CRED_DIR}.tmp"
+    rm -Rf ${temp_dir} && mkdir -p ${temp_dir}
+    if ! unzip device.zip -d ${temp_dir} >/dev/null; then
+	rm -rf ${temp_dir}
         exit_error "Failed extracting credentials file"
     fi
+    sync
 
-    rm -rf $SOTA_BASE_DIR/sql.db
+    rm -rf ${SOTA_CRED_DIR}
+    mv -f ${temp_dir} ${SOTA_CRED_DIR}
 
-    rm -rf $CONFIG_FILE
+    rm -rf ${SOTA_BASE_DIR}/sql.db
+    rm -rf ${CONFIG_FILE}
+    sync
 }
 
 restart_services() {


EdTheBearded:
Hey @jsrc27!
It seems that this sync would help on this scenario where the power cut happens because the operator is waiting for a prompt to power off the device.
But thinking broadly on improving this for power cuts, couldn't we add some hash verification on those files on `check_provisioning_status`? So that we know that the files are there and are correct? Maybe even add a 'flag' file after we confirm this (on the next boot), just so we're not comparing hashes everytime?

Something like:
```
for i in $(unzip -Z1 credentials.zip); do
    orig=$(unzip -p credentials.zip $i | md5sum | cut -d' ' -f1)
    [ -f "$i" ] && file=$(md5sum $i | cut -d' ' -f1) || file=""
    [ "$orig" != "$file" ] && return
done
```

jsrc27:
@EdTheBearded Not sure if I fully understood your feedback. I have some questions for clarification.

First of all did you mean `device.zip` instead of `credentials.zip`? We don't use any `credentials.zip` anywhere in this script.

Also did you mean to suggest adding this logic to `write_credentials` instead of `check_provisioning_status`? Because in `check_provisioning_status` the `device.zip` hasn't been downloaded yet from the server. Or did I misunderstand what you meant?

EdTheBearded:
> @EdTheBearded Not sure if I fully understood your feedback. I have some questions for clarification.
> 
> First of all did you mean `device.zip` instead of `credentials.zip`? We don't use any `credentials.zip` anywhere in this script.
> 
> Also did you mean to suggest adding this logic to `write_credentials` instead of `check_provisioning_status`? Because in `check_provisioning_status` the `device.zip` hasn't been downloaded yet from the server. Or did I misunderstand what you meant?

Yeah, I meant `device.zip`. I wrote with `credentials.zip` because that was a zip file I had in hand to test the output.

No, I meant on `check_provisioning_status`, since this is what decides whether the script will run or not. In the case that we have the zip file and the hashes match, we exit early, if not we just keep running and (re)do the provisioning.

But as @rborn-tx suggested, maybe doing all this in a .tmp folder and as a last step renaming it is a bit cleaner.

jsrc27:
> No, I meant on check_provisioning_status, since this is what decides whether the script will run or not. In the case that we have the zip file and the hashes match, we exit early, if not we just keep running and (re)do the provisioning.

I don't think that would ever realistically be the case no? The `device.zip` gets downloaded during `register_device` into a temp directory in `/tmp`. If the device rebooted then `/tmp` would have been cleared and the zip file would be lost.

For `device.zip` to exist during `check_provisioning_status`, the service would need to have been manually started/restarted while `device.zip` is still in `/tmp` somehow. Which probably means someone is just messing with the process while it's occurring, which is difficult to account for.

So we'd be checking for the case where `device.zip` somehow exists and `auto-provisioning.json` still exists and was not removed from a prior successful run of auto-provisioning. Which would be weird cause how do we have existing certs and `auto-provisioning.json` still exists, but we have cert files installed on the device that do not match what we have in `device.zip`. I just have a hard time imagining the scenario this would be checking for. 

It would make sense if we want to do the hash verification after we unzip just for extra robustness I guess.

EdTheBearded:
> > No, I meant on check_provisioning_status, since this is what decides whether the script will run or not. In the case that we have the zip file and the hashes match, we exit early, if not we just keep running and (re)do the provisioning.
> 
> I don't think that would ever realistically be the case no? The `device.zip` gets downloaded during `register_device` into a temp directory in `/tmp`. If the device rebooted then `/tmp` would have been cleared and the zip file would be lost.
> 
> For `device.zip` to exist during `check_provisioning_status`, the service would need to have been manually started/restarted while `device.zip` is still in `/tmp` somehow. Which probably means someone is just messing with the process while it's occurring, which is difficult to account for.
> 
> So we'd be checking for the case where `device.zip` somehow exists and `auto-provisioning.json` still exists and was not removed from a prior successful run of auto-provisioning. Which would be weird cause how do we have existing certs and `auto-provisioning.json` still exists, but we have cert files installed on the device that do not match what we have in `device.zip`. I just have a hard time imagining the scenario this would be checking for.
> 
> It would make sense if we want to do the hash verification after we unzip just for extra robustness I guess.

Or we download `device.zip` somewhere else. I did not provide a full solution on my comment, just an idea and an example code of hash verification.
I'm certain a bunch of other details would need to be sorted as well.

But either way, @rborn-tx suggestion makes more sense.

jsrc27:
V2 submitted

EdTheBearded (dismissed):

recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
```diff
@@ -136,16 +136,33 @@ register_device() {
...
+
+    # Double-check hashes for extra validation
+    for file in $(unzip -l device.zip | awk '/-----/ {p = ++p % 2; next} p {print $NF}'); do
+        src_hash=$(unzip -p device.zip ${file} | sha256sum | cut -d ' ' -f1)
+	actual_hash=$(sha256sum ${temp_dir}/${file} | cut -d ' ' -f1)
+	if [ "$src_hash" != "$actual_hash" ]; then
+	    rm -rf ${temp_dir}
+            exit_error "Hash mismatch on file $file"
+        fi
+    done
```
rborn-tx, line 158: @jsrc27 I see this came from @EdTheBearded's proposal, but thinking a bit further I couldn't understand how this check would improve resilience to power cuts. Could anyone clarify?

Anyways, let me take the opportunity to explain why I think the temporary directory alone should solve the power cut case.

We have two steps:

1. Unzip multiple files into a temporary directory: This operation can take multiple seconds to complete depending on system load. It would normally involve writing to different areas of the storage device corresponding to the file contents plus some area where the subdirectory entries of `/var/sota` are kept.
2. Rename the directory to the name actually used by the system: This operation can also take a while to start but once started at the filesystem level it would involve just updating a directory name.

If a power cut happens during step 1, the files and/or their file names might be partially written but since they are all inside a temporary directory they wouldn't be used by anyone. So, a power cut during that period shouldn't cause issues with the application. Of course the filesystem itself could be damaged but that's not something we can do anything about at the application level.

As for step 2, the power cut could happen before the system decides to update the directory entry (same case as before), just after it (work already done so not a problem) or during the (re)writing of directory entries. Only the third case would possibly be a problem but that low-level operation is supposed to be very fast. For eMMC, speed is in the order of several MB/s so the write operation should be measured in microseconds; also, I believe the wear leveling would play in our favor.

With this, resilience to power cut should be pretty good unless I'm missing something...

rborn-tx (commented):
Raising some other points.

recipes-sota/auto-provisioning/auto-provisioning/auto-provisioning.sh
jsrc27: Honestly I don't really understand how the extra hash validation would really help either. I was just trying to find a compromise with Ed's suggestion.

EdTheBearded: No, the hash thing was a proposed solution, then @rborn-tx suggested a better one.
Really no need to back to it, just doing as @rborn-tx is plenty fine!

jsrc27:
V3 submitted

rborn-tx (approved):

rborn-tx:
@jsrc27 Would you like to update the commit message before merging this?

jsrc27:
Commit message updated

PR #218, opened 2025-01-30
Add support for Renesas RZ/V2L board

EdTheBearded:


EdTheBearded:
@jsrc27, @rborn-tx, we can finally go ahead with this PR review!

jsrc27 (dismissed):

EdTheBearded:
@rborn-tx, can we merge this?

dynamic-layers/meta-rz-bsp/recipes-bsp/u-boot/u-boot-renesas/0001-include-configs-smarc-rzv2l-add-required-Torizon-boo.patch
```diff
@@ -0,0 +1,45 @@
...
++	"bootcmd_check=if mmc dev 1; then run sd1load; else run emmcload; fi \0" \
++	"devtype=mmc\0" \
++	"devnum=1\0" \
++	"fdtfile=r9a07g054l2-smarc.dtb\0" \
++	"fdt_addr_r=0x50200000\0" \
++	"kernel_addr_load=0x48200000\0" \
++	"kernel_addr_r=0x48200000\0" \
++	"ramdisk_addr_r=0x50300000\0"
++
++#define CONFIG_BOOTCOMMAND	"env default -a;ext4load ${devtype} ${devnum}:1 ${loadaddr} /boot.scr; source"
```
rborn-tx, line 39: Isn't this going to throw away all the changes to the U-Boot environment on every boot? If so, it this intended? Also, one suggestion I gave a former PR from @lucas-akira was that we could make the source command conditional on ext4load being successful: `ext4load ... && source`.

rborn-tx (commented):
Sorry for the delay. Just raising a question for clarification and adding a suggestion...

dynamic-layers/meta-rz-bsp/recipes-bsp/u-boot/u-boot-renesas/0001-include-configs-smarc-rzv2l-add-required-Torizon-boo.patch
rborn-tx: Resetting the environment would break functionalities requiring a persistent environment, such as the rollback feature. I'm not sure how the U-Boot initial environment is set on this platform, but any changes to the initial environment would be lost as well (since U-Boot would reset variables to its internal/hard-coded values).

Is that what we want here?

EdTheBearded: You mean the counting variables to trigger fallback?
If so, then what should we do here?
My original solution was without resetting the environment, but then when someone from Renesas tested, our new environment wasn't present (because it was loading the old one that was saved). So as a way around it I changed my commit to the way it was done upstream:
https://github.com/renesas-rz/renesas-u-boot-cip/blob/v2021.10/rz/include/configs/smarc-rzv2l.h#L72

The only way I see here is tor remove this `env default` and then the user would be required to run this manually the first time the board is flashed.

rborn-tx: Hmm, good question. The initial setup of the environment is normally done by an installer of some sort. But here I imagine we don't have one, right?

I recall the Android EVK from NXP had instructions asking the user to do this initial setup manually. In that case the `bootcmd` variable would be empty and U-Boot would stop at its CLI. The user would then run some commands maybe like this:

```
U-Boot# env default -a
U-Boot# env set bootcmd 'ext4load ${devtype} ${devnum}:1 ${loadaddr} /boot.scr && source'
U-Boot# env save
```

How about doing something similar?

rborn-tx: Sounds like a plan :grin:

EdTheBearded: Awesome, done!

The docs I'll be adding on another PR, along with the required setup script.

rborn-tx (approved):

jsrc27 (approved):

PR #220, opened 2025-02-03
imx95: install mali firmware (mali_csffw.bin) in the rootfs

leonheldattoradex:
We need the firmware as the kernel module expects it in specific places of the
root filesystem, so it cannot be dynamically deployed from containers as the
rest of the graphical stack is.
    
On the imx8 and am62 for example the firmware is installed via 'driver' recipes
(kernel-module-imx-gpu-viv and ti-img-rogue-installer), on the imx95, however,
it comes from userspace portion, thus we need the bbappend to only install the
firmware, nothing else (no libEGL, libGLES etc...).
    
This follows the same logic as commits a7ec4b08 and b40b9114.

Note that I haven't 100% tested this commit besides a build as I need to actually build the graphical userspace to trigger the kernel module to load the firmware, but having it in `lib/firmware` is, from my understanding, enough.

Of course, the rest of the graphical stack belongs inside one of these: https://github.com/torizon/torizon-containers.

### DIFF
--- a/recipes-images/images/torizon-base.inc
+++ b/recipes-images/images/torizon-base.inc
@@ -80,6 +80,11 @@ CORE_IMAGE_BASE_INSTALL:append:sota = " \
     tzn-mqtt \
 "
 
+# Packages that install firmwares/kernel modules that must be present
+# in the rootfs. Firmware blobs, generally speaking, cannot be loaded
+# from containers as their kernel module counterparts generally expect
+# the files to be in an specific directory (generally /lib/firmware)
+
 CORE_IMAGE_BASE_INSTALL:append:mx8-nxp-bsp = " \
     kernel-module-imx-gpu-viv \
 "
@@ -90,6 +95,10 @@ CORE_IMAGE_BASE_INSTALL:append:am62xx = " \
 
 CORE_IMAGE_BASE_INSTALL:append:mx8mp-nxp-bsp = " \
     kernel-module-isp-vvcam \
+"
+
+CORE_IMAGE_BASE_INSTALL:append:mx95-nxp-bsp = " \
+    mali-imx \
 "
 
 nss_altfiles_set_users_groups () {
--- a/dynamic-layers/meta-imx-bsp/recipes-graphics/mali/mali-imx_%.bbappend
+++ b/dynamic-layers/meta-imx-bsp/recipes-graphics/mali/mali-imx_%.bbappend
@@ -1,0 +1,7 @@
+DEPENDS = ""
+
+do_install () {
+    install -d ${D}${nonarch_base_libdir}/firmware
+    cp -r ${S}/usr/lib/firmware/* ${D}${nonarch_base_libdir}/firmware
+}
+


recipes-graphics/mali/mali-imx_r50.0.bbappend
EdTheBearded, line 1: @leonheldattoradex, this seems to be an .bbappend for the meta-imx-bsp layer, right? I believe it would be better if we added this .bbappend into dynamic-layers.

recipes-images/images/torizon-base.inc
EdTheBearded, line 1: Instead of changing `torizon-base.inc`, I think would be better organized adding the `CORE_IMAGE_BASE_INSTALL:append` directly into the machine file `conf/machine/include/imx95-19x19-verdin.inc`.

recipes-graphics/mali/mali-imx_r50.0.bbappend
leonheldattoradex: It's for `meta-freescale`, I'm assuming we're building from that, right?

recipes-images/images/torizon-base.inc
EdTheBearded: No.
It's just that when we talk about common machines, I think i'ts best if any things needed don't spill into core Torizon stuff.

recipes-graphics/mali/mali-imx_r50.0.bbappend
EdTheBearded: Both layers have this recipe, don't know now from which layer we're actually getting it. Probably an easy way to check would be to run `bitbake mali-imx -c devshell` and comparing the checkout hash with the recipes

recipes-images/images/torizon-base.inc
leonheldattoradex: Gotcha. I'm thinking that eventually we'll have smarc95 (and others) as a machine which will be `COMPATIBLE` with `mx95-nxp-bsp`. ie, the am62 fix for this same issue works for all am62 models, not only for an specific machine. What you think?

EdTheBearded: Hmm, that makes sense. We can keep this in `torizon-base.inc`.

recipes-graphics/mali/mali-imx_r50.0.bbappend
leonheldattoradex: Okay maybe I did something extremely wrong, was there supposed to be a `meta-imx` in the layers? I just `repo sync`, set the machine, distro and `bitbake`, no `meta-imx` on sight.

recipes-images/images/torizon-base.inc
leonheldattoradex: :catjam:

recipes-graphics/mali/mali-imx_r50.0.bbappend
EdTheBearded: Yup, there's no manifest for common-machines yet, so there's no way to `repo sync` iMX95. :grinning:
If you want to sync and have a local build for this just let me know and I can share with you the steps.

leonheldattoradex:
Alright drafting for a bit as I wasn't building it correctly.

recipes-graphics/mali/mali-imx_r50.0.bbappend
EdTheBearded: Also, aside from adding this to dynamic-layers, we should not fix the version of this .bbappend, and rather rename the file to `mali-imx_%.bbappend`.

leonheldattoradex: Done!

leonheldattoradex:
Verified it works

```
torizon@localhost:~$ ls /lib/firmware/
LICENCE.Marvell  LICENCE.rtlwifi_firmware.txt  mali_csffw.bin  mrvl  regulatory.db  regulatory.db.p7s  rtlwifi	wave633c_codec_fw.bin
torizon@localhost:~$ ls /usr/lib/firmware/
LICENCE.Marvell  LICENCE.rtlwifi_firmware.txt  mali_csffw.bin  mrvl  regulatory.db  regulatory.db.p7s  rtlwifi	wave633c_codec_fw.bin
```

recipes-graphics/mali/mali-imx_r50.0.bbappend
leonheldattoradex: oh wait you asked me to move it to the dynamic layers.

Hang on a sec.

leonheldattoradex:
Also tested with https://github.com/torizon/imx-gpu-mali, works fine:

```
[  164.557996] mali 4d900000.gpu: Loading Mali firmware 0x1010000
[  164.559797] mali 4d900000.gpu: Protected memory allocator not found, Firmware protected mode entry will not be supported
[  164.571192] mali 4d900000.gpu: Mali firmware git_sha: 8e5cfcfec20cc8aff8509d37e72babc935d34a3b 
[  164.573699] mali 4d900000.gpu: CSF_SCHED_PROTM_PROGRESS_TIMEOUT is capped from 20971ms to 4500ms
[  164.575944] mali 4d900000.gpu: Clearing BASE_MEM_UNCACHED_GPU flag to avoid MMA violation
```

jsrc27 (approved):

EdTheBearded (approved):

PR #224, opened 2025-02-05
iMX95: fix some build issues

EdTheBearded:
iMX95 was not booting properly, because the uEnv.txt present in the image was not the one properly for Common Torizon.

Also fixed the bootloader, adding the fragments to enable rollback and re-enabled 'OSTREE_DEPLOY_DEVICETREE' for iMX95, so that it's dtb and dtbo's are deployed into the device.

Related-to: TOR-3752


jsrc27 (approved):

rborn-tx (approved):
LGTM

PR #225, opened 2025-02-05
auto-provisioning: improve behavior with power cuts (kirkstone)

jsrc27:
Some users when auto-provisioning, power down the system once they see the auto-provisioning service exit successfully. However, in some cases this can cause the files created from the provisioning process to not be fully written to disk, resulting in 0-byte files upon next boot.

Therefore, add some explicit syncs to improve this behavior. Also make use of a temp directory to improve the overall atomicity of the process.

Related-to: TOR-3743

Sames as: https://github.com/torizon/meta-toradex-torizon/pull/217 but for Kirkstone


rborn-tx (approved):
LGTM

EdTheBearded (approved):

PR #229, opened 2025-02-07
initramfs-framework: fix torizon-signed SPDX build issue

rborn-tx:
This is expected to solve a build issue with torizon-signed images where the following error message is output in the build logs:

ERROR: initramfs-ostree-torizon-image-1.0-r0 do_rootfs: Cannot find any SPDX file for document
http://spdx.org/spdxdocs/kernel-module-display-connector-<version>-<sha>

The immediate cause for the problem comes from the initramfs-framework: do_create_runtime_spdx task not being executed even when initial ramdisk undergoes modifications between two builds. Unfortunately the root cause is still a mistery. However, a simple solution employed here is by making the initramfs-framework:do_create_runtime_spdx task "rdepend" on the corresponding task of the kernel.

Besides the above, also change how we set the dependencies to the required kernel modules when composefs is enabled. Initially the dependencies were set as initramfs-module-composefs -> kernel modules but now the dependency is set as initramfs-module-kmod -> kernel modules as done with all the other modules. Tests have shown that this wasn't the root cause of the SPDX issue though.

Related-to: TOR-3704


EdTheBearded (approved):
Nice!!!
All that because of unmanaged dependencies!!!

rborn-tx:
> All that because of unmanaged dependencies!!!

I hope so :grinning:

jsrc27 (approved):

PR #231, opened 2025-02-10
aktualizr-torizon: bump SRCREV to latest

rborn-tx:
Bump the `aktualizr-torizon` version being used (just to synchronize with TorizonCore Builder).


jsrc27 (approved):

EdTheBearded (approved):
There's no problem in bumping the hash, but there's also no real effect outside of a monthly or quarterly release, since our nightlies are ausing `AUTOREV`.
And before every release we also have a task to bump the hashes on our recipes.

PR #236, opened 2025-02-12
rac: enable RAC by default

EdTheBearded:
Now that it is using MQTT, data usage is more tolerable and we can have it enabled by default on Torizon OS builds.

Related-to: TOR-3750

### DIFF
--- a/recipes-sota/rac/rac_git.bb
+++ b/recipes-sota/rac/rac_git.bb
@@ -29,8 +29,6 @@ SRCREV_tough = "9316c096b32196df75ba17a8a5502b19baffe24e"
 S = "${WORKDIR}/git"
 
 SYSTEMD_SERVICE:${PN} = "remote-access.service"
-# Keep disabled by default for now
-SYSTEMD_AUTO_ENABLE:${PN} = "disable"
 
 PV = "0.0+git${SRCPV}"
 
--- a/recipes-sota/tzn-mqtt/tzn-mqtt_git.bb
+++ b/recipes-sota/tzn-mqtt/tzn-mqtt_git.bb
@@ -21,8 +21,6 @@ LIC_FILES_CHKSUM = " \
 "
 
 SYSTEMD_SERVICE:${PN} = "tzn-mqtt.service"
-# Keep disabled by default for now
-SYSTEMD_AUTO_ENABLE:${PN} = "disable"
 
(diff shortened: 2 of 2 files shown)


recipes-sota/rac/rac_git.bb
```diff
@@ -29,8 +29,7 @@ SRCREV_tough = "9316c096b32196df75ba17a8a5502b19baffe24e"
 S = "${WORKDIR}/git"
 
 SYSTEMD_SERVICE:${PN} = "remote-access.service"
-# Keep disabled by default for now
-SYSTEMD_AUTO_ENABLE:${PN} = "disable"
+SYSTEMD_AUTO_ENABLE:${PN} = "enable"
```
jsrc27, line 32: I think we can just drop this line completely right? Since `SYSTEMD_AUTO_ENABLE` should default to enable if I recall.

EdTheBearded: It does, but I just thought that having it explicitly enabled would make things clear.
I can remove the line if you want

jsrc27: Well I don't have a strong preference. I was just wondering if there was a reason you went with the explicit route, since we usually don't in our other recipes.

jsrc27:
For the MQTT stuff to work on RAC do we also need to enable the tzn-mqtt service from this recipe as well?: https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/recipes-sota/tzn-mqtt/tzn-mqtt_git.bb

Or is that a completely separate thing?

recipes-sota/rac/rac_git.bb
EdTheBearded: I've removed the `enable` line.

EdTheBearded:
> For the MQTT stuff to work on RAC do we also need to enable the tzn-mqtt service from this recipe as well?: https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/recipes-sota/tzn-mqtt/tzn-mqtt_git.bb
> 
> Or is that a completely separate thing?

Not necessarily, the service would remain disabled, but the dependency chain of RAC would start MQTT.
But for clarity I also enable `tzn-mqtt`.

rborn-tx (approved):
LGTM

PR #239, opened 2025-02-17
Add support for Xenomai on Toradex SoMs

leograba:
This PR adds support for building the Linux kernel based on version 6.12 with the Xenomai patches, both for Xenomai 3 and 4.

It makes it in a way that the recipe also accommodate the x86 machine intel-corei7-64, which already has support for building Xenomai on top of the kernel versions 5.4 and 5.10.

The biggest caveat for the moment is that 6.12 is not supported by meta-intel on scarthgap, therefore there are overrides to build 6.6 for intel-corei7-64. I initially did it this way so it's easy to drop these overrides on master and on upcoming versions of Yocto. Also, it may be possible to backport the 6.12 Intel kernel and drop the overrides.

Like other variants of software that can be built for Toradex hardware (Podman and RT are examples), these Xenomai distros should be considered experimental.


leograba:
Adding this PR as a draft because I still have to do some runtime tests especially to check the kernel command-line arguments, but I already would like to start getting feedback on the Yocto part of things.

conf/distro/torizon-upstream-xenomai3.conf
EdTheBearded, line 1: This distro is building upstream kernel for all devices, or is only for the devices we use upstream kernel? I understand it's the first, right? Since `upstream` DISTROOVERRIDES is being added.

Couldn't we just name these `torizon-xenomai[3|4]`?

leograba: > This distro is building upstream kernel for all devices, or is only for the devices we use upstream kernel? I understand it's the first, right? Since upstream DISTROOVERRIDES is being added.

For intel-corei7-64, the upstream override does not apply. This is why I had to do this:

```
PREFERRED_PROVIDER_virtual/kernel:xenomai4 = "linux-xenomai-4"
PREFERRED_PROVIDER_virtual/kernel:xenomai4:upstream = "linux-xenomai-4"
```

For Toradex machines, I'm explicitly using the upstream kernel recipe as base:

```
# Use BSP kernel recipe as the base
BASE_KERNEL_CONFIG:aarch64:tdx = "recipes-kernel/linux/linux-toradex-upstream_6.12.bb"
```

In the end, I could try to actually not use the `upstream` override for Toradex BSPs, even if I use the linux-toradex-upstream. But then we come to your second question, which is more conceptual than practical.

> Couldn't we just name these torizon-xenomai[3|4]?

These Xenomai kernels aren't technically upstream because they are a fork of the upstream + the Xenomai patches applied:

- https://source.denx.de/Xenomai/xenomai4/linux-evl (Xenomai 4)
- https://source.denx.de/Xenomai/linux-dovetail + https://source.denx.de/Xenomai/xenomai (Xenomai 3)

But, from our perspective, what we call downstream are the SoC vendor BSPs, at least so far. And these Xenomai are not based on the SoC vendor BSPs. They are "just" some patches on top of upstream kernel.

We have PREEMPT_RT variants of the upstream kernel, which looks more similar to this Xenoami case than a downstream.

And we still called those `linux-toradex-upstream-rt` https://git.toradex.com/cgit/meta-toradex-bsp-common.git/tree/recipes-kernel/linux/linux-toradex-upstream-rt_6.6.bb?h=scarthgap-7.x.y - or at least we did before PREEMPT_RT was merged into the upstream.

I mean, whatever you guys decide, I'm in. I don't have a strong preference.

leograba: p.s. would the question apply the other way around? I mean, for x86 the Xenomai recipe is using the "upstream" kernel plus the patches, instead of the linux-intel which uses github.com/intel/linux-intel-lts

I don't know if we can consider this kernel Intel's downstream, or just a fork of the upstream.

But if we consider it Intel's downstream, I think that for the sake of alignment, then we should have:

- Either: set `torizon-upstream-xenomai` and rename `s/common-torizon-xenomai/common-torizon-upstream-xenomai/`
- Or: set `torizon-xenomai` and keep `common-torizon-xenomai`

recipes-kernel/linux/linux-xenomai-3_6.12.bb
```diff
@@ -0,0 +1,82 @@
+SUMMARY = "Linux kernel with Xenomai 3 Dovetail and the Cobalt core patches"
```
jsrc27, line 1: Minor point. Could some of the common aspects of this file and `linux-xenomai-4_6.12.bb` be factored out to some common include file?

leograba: Sure. I'll put some thinking into it and come back with a proposal

conf/distro/torizon-upstream-xenomai3.conf
EdTheBearded: I think I understood your reasoning!
I like your suggestions. As I'm a sucker for short names, I would love to see option 2, but I think option 1 is perfectly fine as well.
@jsrc27, @rborn-tx, any opinions on this?

rborn-tx: I believe that for our SoMs the idea of adding a -upstream suffix was to allow customers to select between upstream or downstream (whatever that means), not to inform users if this is an upstream or downstream "distro".

I wrote "whatever that means" because in the past (e.g. kirkstone) downstream meant using both kernel and bootloader from forks we maintain (or co-maintain) along with everything that comes with them; and upstream meant using kernel and bootloader from their original sources (kernel.org and denx, respectively). With scarthgap the bootloader may come from denx or from our forks depending on the machine so that the upstream/downstream terms no longer apply to the bootloader. So, at the moment, if we interpret that upstream indicates the origin of the kernel, then the term upstream becomes clear, at least for our SoMs.

Well, I wrote all of the above just to say that for our SoMs I believe we should follow the existing naming pattern because the term "upstream" is still meaningful; in which case the name being proposed by Leo here of "torizon-upstream-xenomai3" looks good to me.

We might consider though basing the distro file on `torizon-upstream.conf` instead because the ":upstream" override would be added on a single place. Notice, I'm not saying this is better - I'm just raising the topic for discussion.

Regarding common-torizon, I don't think we need to follow the same pattern because there even the idea of upstream/downstream could be blurry. And as I said before, the idea of having the -upstream suffix is allowing people to switch, not to inform them. So, for common-torizon, I don't see the need for changing names. Actually, if we ever have multiple possibilities for the provider of the kernel, then I'd name the distro files with something more specific like `common-torizon-intel-xenomai` or something like that.

Makes sense?

conf/distro/include/xenomai.inc
```diff
@@ -0,0 +1,22 @@
...
+XENOMAI_KERNEL_ARGS = "audit=0 clocksource=arch_sys_counter nohz_full=1,3 \
+                       irqaffinity=0 isolcpus=1,3 rcu_nocbs=1,3 rcu_nocb_poll=1024 \
+                       nosoftlockup numa_balancing=disable hugepages=1024 "
+OSTREE_KERNEL_ARGS:sota:append:xenomai4 = " ${XENOMAI_KERNEL_ARGS} "
+OSTREE_KERNEL_ARGS:sota:append:xenomai3 = " ${XENOMAI_KERNEL_ARGS} "
+
+PREFERRED_PROVIDER_virtual/kernel:xenomai4 = "linux-xenomai-4"
+PREFERRED_PROVIDER_virtual/kernel:xenomai4:upstream = "linux-xenomai-4"
+PREFERRED_PROVIDER_virtual/dtb:xenomai4:upstream = "device-tree-overlays-mainline"
+LINUX_VERSION_EXTENSION:xenomai4:upstream ?= "-torizon-${LINUX_KERNEL_TYPE}-xenomai4"
```
rborn-tx, line 13: Could you clarify why you need both `:xenomai4` and `:upstream` overrides on these assignments? Looking at the distro config both overrides are added together so either override would do the job, no?

However, using one or the other might make more sense: in this stretch of code I'd imagine something like this:

```
PREFERRED_PROVIDER_virtual/kernel:xenomai4 = "linux-xenomai-4"
PREFERRED_PROVIDER_virtual/dtb:upstream = "device-tree-overlays-mainline"
LINUX_VERSION_EXTENSION:xenomai4 ?= "-torizon-${LINUX_KERNEL_TYPE}-xenomai4"
```

The xenomai specific assignments use `:xenomai4` while the DTB provider selection depends on `:upstream`, not on the fact that we're using xenomai. Doesn't this work?

rborn-tx (commented):
Good work! I've raised some points for discussion.

conf/distro/torizon-upstream-xenomai3.conf
rborn-tx: I'd try to keep the naming pattern at least for distros targeting our SoMs; so I'd go with `torizon-upstream-xenomai`.

For common-torizon I don't have any preference. If we don't foresee the possibility of providing both downstream and upstream distros or if the meaning of these terms is too blurry I wouldn't necessarily use those words in the distro file name. Thus even the current `common-torizon-xenomai` would be okay.

rborn-tx:
@leograba Could we drop this PR?

PR #240, opened 2025-02-18
Add TI AM62x Starter Kit EVM documentation and setup script

lucas-akira:


lucas-akira:
Pinging @EdTheBearded @jsrc27 @rborn-tx here since I cannot assign reviewers to the PR directly.

jsrc27 (dismissed):

docs/README-ti.md
```diff
@@ -0,0 +1,39 @@
...
+> Common Torizon OS is only available on branches `scarthgap-7.x.y` or newer!
+
+Build
+======
+1. Source `setup-environment` to setup build, specifying the machine to build with the MACHINE variable e.g.:
+```bash
+$ MACHINE=am62xx-evm . setup-environment <build-directory>
+```
+If a build directory is not given, the script will create one named `build-ti-${DISTRO}`, where all build artifacts will be stored. By default `DISTRO` is automatically set to `common-torizon` with the TI boards.
+
```
EdTheBearded, line 22: We're going with no blankspaces between items

docs/README-ti.md
```diff
@@ -0,0 +1,39 @@
+Sync
```
EdTheBearded, line 1: Could you please also add a section on how top syn "manually"?
By using git, cloning their layers and then `meta-toradex-torizon` and its dependencies?

docs/README-ti.md
```diff
@@ -0,0 +1,39 @@
...
+2. Inside the build directory, start the build e.g.:
+```bash
+$ bitbake torizon-docker
+```
+
+All artifacts should be inside `<build-directory>/deploy/images/${MACHINE}`, including the `.wic` file.
+
+Booting the image from an SD card
+======
+1. Flash the generated .wic artifact file to a micro SD card and insert it into the board.
```
EdTheBearded, line 32: For completeness, could you just add the `dd` command here for copying the .wic image?

docs/README-ti.md
```diff
@@ -0,0 +1,39 @@
...
+
```
EdTheBearded, line 33: Blank space

docs/README-ti.md
```diff
@@ -0,0 +1,39 @@
...
+2. Configure the board for SD card boot and connect the UART interface to the host PC. For example, for the AM62x SK EVM you can follow the instructions on the [TI Resource Explorer website](https://dev.ti.com/tirex/explore/node?node=A__AZatcAgHClq18snt16Hmfg__PROCESSORS-DEVTOOLS__FUz-xrs__LATEST).
+
```
EdTheBearded, line 35: Blank space

scripts/lib/setup-devices/setup-environment-ti
```diff
@@ -0,0 +1,133 @@
...
+LCHAIN_PATH_ARMV8 EXTRA_TISDK_FILES TISDK_VERSION ARAGO_BRAND ARAGO_RT_ENABLE ARAGO_SYSTEST_ENABLE ARAGO_KERNEL_SUFFIX TI_SECURE_DEV_PKG_CAT TI_SECURE_DEV_PKG_AUTO TI_SECURE_DEV_PKG_K3 ARAGO_SYSVINIT SY
+SFW_FILE"
+EOF
+
+if [ -z "${MACHINE}" ]; then
+  echo "MACHINE not set. Defaulting to am62xx-evm."
+  MACHINE="am62xx-evm"
+fi
+
+if [ -z "${DISTRO}" ]; then
```
EdTheBearded, line 74: I believe we can remove this, since we're always calling this script with DISTRO set to common-torizon

scripts/lib/setup-devices/setup-environment-ti
```diff
@@ -0,0 +1,133 @@
...
+unset SCRIPTS
+export PATH
+export BUILDDIR="${BUILDDIR}"
+export BB_ENV_PASSTHROUGH_ADDITIONS="MACHINE DISTRO TCMODE TCLIBC http_proxy ftp_proxy https_proxy all_proxy ALL_PROXY no_proxy SSH_AGENT_PID SSH_AUTH_SOCK BB_SRCREV_POLICY SDKMACHINE BB_NUMBER_THREADS
+PARALLEL_MAKE GIT_PROXY_COMMAND GIT_PROXY_IGNORE SOCKS5_PASSWD SOCKS5_USER OEBASE META_SDK_PATH TOOLCHAIN_TYPE TOOLCHAIN_BRAND TOOLCHAIN_BASE TOOLCHAIN_PATH TOOLCHAIN_PATH_ARMV5 TOOLCHAIN_PATH_ARMV7 TOO
+LCHAIN_PATH_ARMV8 EXTRA_TISDK_FILES TISDK_VERSION ARAGO_BRAND ARAGO_RT_ENABLE ARAGO_SYSTEST_ENABLE ARAGO_KERNEL_SUFFIX TI_SECURE_DEV_PKG_CAT TI_SECURE_DEV_PKG_AUTO TI_SECURE_DEV_PKG_K3 ARAGO_SYSVINIT SY
+SFW_FILE"
+EOF
+
+if [ -z "${MACHINE}" ]; then
```
EdTheBearded, line 69: I don't think we should have a default machine here.
Since we have a menu for selection, if no machine is passed, we won't even reach this specific script.
And furthermore, this will be shared with other TI devices, so I believe it is safe to just assume that, if we're here, we have both MACHINE and DISTRO set

scripts/lib/setup-devices/setup-environment-ti
```diff
@@ -0,0 +1,133 @@
...
+  cat <<EOF >>conf/local.conf
+# Where to save the packages and images
+TI_COMMON_DEPLOY = "\${TOPDIR}/deploy"
+
+# Torizon
+include conf/machine/include/\${MACHINE}.inc
+EOF
+fi
+
+if ! grep -q "# Torizon" conf/bblayers.conf; then
```
EdTheBearded, line 101: Maybe we could move this to the block on line 9?
Since it is already appending to bblayers.conf.

docs/README-ti.md
lucas-akira: I can remove the other blank lines but we need this one, as otherwise there won't be a line break between this last sentence:
 > ...is automatically set to `common-torizon` with the TI boards.

and the following one:

> 2. Inside the build directory...


At least the GitHub markdown preview showed the result like this without the blank line:

> If a build directory is not given, the script will create one named build-ti-${DISTRO}, where all build artifacts will be stored. By default DISTRO is automatically set to common-torizon with the TI boards. 2. Inside the build directory, start the build e.g.:

scripts/lib/setup-devices/setup-environment-ti
lucas-akira: Yeah, I put default values for machine and distro as a safeguard just in case the script doesn't have those variables set for whatever reason.

But I agree that it makes more sense for the caller in `setup-environment` to properly take care of that. I'll remove this and the DISTRO check.

docs/README-ti.md
lucas-akira: Sure, I'll add it.

docs/README-ti.md
lucas-akira: Huh, I had no idea that adding 2 trailing whitespaces produces a line break.

But given that we're creating a separation between items of a numbered list, I think a new paragraph space (produced by the blank line) is more suitable compared to a line break produced by the two trailing whitespaces.

With a regular line break the next item of the list is too close to the previous one, being hard to know when one ends and the other one starts:

image

With the new paragraph we have this:

image

What do you think?

EdTheBearded: This was a doc formatting thing, that @leograba had pointed out.
Now I'm really unsure about this. Leo, is there a coding style page for markdown somewhere that you follow?

docs/README-ti.md
lucas-akira: Done!

docs/README-ti.md
lucas-akira: Thanks, I removed it.

docs/README-ti.md
lucas-akira: Solved.

docs/README-ti.md
lucas-akira: Alright, I've renamed the section with the `repo` commands to `Repo Sync` and added a new section where everything is manually cloned called `Sync`.

scripts/lib/setup-devices/setup-environment-ti
lucas-akira: OK, done!

scripts/lib/setup-devices/setup-environment-ti
lucas-akira: Done!

docs/README-ti.md
leograba: I give up

I've been the pedantic one before and sort of regretted it after a while

Whatever works, guys

scripts/lib/setup-devices/setup-environment-ti
lucas-akira: If we move this part inside the block on line 9 it would only be executed if `conf/bblayers.conf` doesn't already exist. I put it separately because I wanted this check in particular to always happen when running the script.

EdTheBearded: I think in the end it would produce the same result, but in the end I'm fine either way!

EdTheBearded (approved):

PR #242, opened 2025-02-24
imx95: add torizon.cfg fragment to NXP's Kernel

EdTheBearded:
To apply a Kernel fragment to NXP's Kernel, we need to add it to 'DELTA_KERNEL_DEFCONFIG', so that it can be manually applied to the generated .config file.

Related-to: TOR-3754


jsrc27 (approved):

rborn-tx (approved):
LGTM

PR #243, opened 2025-02-24
setup-environment: discard 'unset' error messages

EdTheBearded:
Sometimes the 'unset' command will output some error that it wasn't able to unset a variable, and will pollute the output we give.

So now we suppress the error output of the command.

Related-to: TOR-3763


jsrc27 (approved):

rborn-tx (approved):
LGTM

PR #252, opened 2025-03-19
am62xx: make bbappend changes only apply to TI's AM62 builds

EdTheBearded:
We have some 'meta-ti-bsp' custom .bbappends on dynamic-layers, but the issue here is that Toradex's HW also uses this layer so we end up apllying those configurations as well.

In the end, it resulted in 'fw_printenv' not working properly, since the configuration file for TI is different from Toradex, which caused greenboot to not work properly and the OS rollback process to fail altogether.

Now we add a new override for TI's AM62 devices in our 'common-am62xx.inc' file, namely 'common-ti', since there's not unique override on TI's BSP, and 'ti-soc' that was used was common to both BSP's. So now, on TI's specific .bbappends, we use the 'common-ti' override, to make sure we're using Common Torizon distro and TI machine.

Related-to: TOR-3777

### DIFF
--- a/conf/machine/include/common-am62xx.inc
+++ b/conf/machine/include/common-am62xx.inc
@@ -1,5 +1,8 @@
 # Base configurations for TI's BSP AM62 devices
 
+# Adding custom override to differentiate between Verdin AM62 and TI's AM62xx EVMs
+MACHINEOVERRIDES =. "common-ti:"
+
 WKS_FILE:sota = "torizon-am62xx-sota.wks"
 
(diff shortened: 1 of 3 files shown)


jsrc27 (approved):
LGTM

By the way, small typo in the commit message:
> we end up _apllying_ those configurations as well.

rborn-tx (approved):
LGTM

EdTheBearded:
> LGTM
> 
> By the way, small typo in the commit message:
> 
> > we end up _apllying_ those configurations as well.

Fixed.

PR #253, opened 2025-03-26
Fix Common Torizon builds for AM62x EVM machine

lucas-akira:
With the latest changes to meta-toradex-torizon, Common Torizon no longer builds successfully for the AM62 EVM.

With the addition of the BeagleY-AI support, the `setup-environment-ti` script always includes the `meta-beagle` layer to `conf/bblayers.conf` for all TI machines. This layer is a fairly recent addition to the `meta-ti` repository.

The problem is that the current `meta-ti` commit that we checkout in our Common Torizon manifest for our AM62x and AM62P EVM builds doesn’t have `meta-beagle` yet, so Bitbake complains that it cannot find the missing layer.

The setup script has been adapted to only add the layer to `bblayers.conf` when building an image for the `beagley-ai` machine, and it removes references to the layer when building other TI machines.

-----

The recent addition of a `common-am62xx.inc` base config file applied for AM62 machines has apparently caused the recipe to use the default value of `KERNEL_DEVICETREE` set by the TI BSP, even with a hard assignment to the variable in the machine include file. Maybe the variable should have an override?

The default value for this variable references an overlay which is not present in the kernel sources. This eventually causes issues in the `ostree-kernel-initramfs` recipe, where it tries to copy a non-existent .dtbo file to the image filesystem.

It would be probably better to not have the hard assignment in our layer in the first place: variables set at the vendor BSP level should be kept untouched whenever possible, as those can change with updates.

One way I found to solve this issue is to hard assign `OSTREE_DEVICETREE` instead of `KERNEL_DEVICETREE`. See more details in the commit messages.

Related-to: TOR-3782

Signed-off-by: Lucas Akira Morishita <lucas.morishita@toradex.com>

### DIFF
--- a/scripts/lib/setup-devices/setup-environment-ti
+++ b/scripts/lib/setup-devices/setup-environment-ti
@@ -18,7 +18,6 @@ BBLAYERS += " \\
     \${OEROOT}/sources/meta-openembedded/meta-filesystems \\
     \${OEROOT}/sources/meta-ti/meta-ti-extras \\
     \${OEROOT}/sources/meta-ti/meta-ti-bsp \\
-    \${OEROOT}/sources/meta-ti/meta-beagle \\
     \${OEROOT}/sources/meta-arm/meta-arm \\
     \${OEROOT}/sources/meta-arm/meta-arm-toolchain \\
     \${OEROOT}/sources/oe-core/meta \\
@@ -72,6 +71,14 @@ source conf/setenv
 # Update DISTRO and MACHINE
 sed -E -i "/(^|#|#.*( |#))DISTRO .*/c\DISTRO = '${DISTRO}'" conf/local.conf
 sed -E -i "/(^|#|#.*( |#))MACHINE .*/c\MACHINE = '${MACHINE}'" conf/local.conf
+
+# Check if meta-beagle should be added to bblayers.conf or not
+if [ "${MACHINE}" = "beagley-ai" ]; then
+  if ! grep -q "^BBLAYERS.*/meta-beagle" conf/bblayers.conf; then
+    cat <<EOF >>conf/bblayers.conf
+BBLAYERS += "\${OEROOT}/sources/meta-ti/meta-beagle"
+EOF
+fi
 
 if ! grep -q "# Torizon" conf/local.conf; then
   echo "Adapting conf/local.conf for a Torizon OS build."


scripts/lib/setup-devices/setup-environment-ti
```diff
@@ -73,6 +72,20 @@ source conf/setenv
...
+  if ! grep -q "meta-beagle" conf/bblayers.conf; then
+    cat <<EOF >>conf/bblayers.conf
+BBLAYERS += "\${OEROOT}/sources/meta-ti/meta-beagle"
+EOF
+  else
+    sed -E -i '/(^#|^#.*( |#))BBLAYERS.*meta-beagle/c\BBLAYERS += "\${OEROOT}\/sources\/meta-ti\/meta-beagle"' conf/bblayers.conf
+  fi
+else
+  # Remove reference to meta-beagle if not building a BeagleY-AI image
+  sed -i '/^BBLAYERS.*\/meta-beagle/d' conf/bblayers.conf
```
EdTheBearded, line 86: I don't think we need to worry about removing meta-beagle if present while building other machines.

scripts/lib/setup-devices/setup-environment-ti
```diff
@@ -73,6 +72,20 @@ source conf/setenv
...
 sed -E -i "/(^|#|#.*( |#))MACHINE .*/c\MACHINE = '${MACHINE}'" conf/local.conf
 
+# Check if meta-beagle should be added to bblayers.conf or not
+if [ "$MACHINE" = "beagley-ai" ]; then
+  if ! grep -q "meta-beagle" conf/bblayers.conf; then
+    cat <<EOF >>conf/bblayers.conf
+BBLAYERS += "\${OEROOT}/sources/meta-ti/meta-beagle"
+EOF
+  else
+    sed -E -i '/(^#|^#.*( |#))BBLAYERS.*meta-beagle/c\BBLAYERS += "\${OEROOT}\/sources\/meta-ti\/meta-beagle"' conf/bblayers.conf
```
EdTheBearded, line 82: Is this to remove a commented out meta-beagle? I think we could also skip this... What was the idea behind this one?

lucas-akira: Right, that's for the situation where MACHINE is set as `beagley-ai` and `bblayers.conf` already has the `meta-beagle` string for whatever reason; maybe the user manually commented the `meta-beagle` inclusion line when building another machine, and now wants to build a BeagleY image, but forgot to uncomment the line.

I tried to imagine edge case scenarios just so the script can still work in these situations.

scripts/lib/setup-devices/setup-environment-ti
lucas-akira: I wanted to have this behavior just to make sure the 'missing layer' issue never happens, even if the user decides to do something unusual, like reusing the same build environment/directory for a BeagleY-AI image to an AM62x EVM image with the "older" `meta-ti` layer that doesn't have `meta-beagle` yet.

lucas-akira: Fair enough, I guess it makes more sense to try to keep things simpler on our side. I'll remove this and the other part.

lucas-akira: OK, done!

EdTheBearded (approved):

scripts/lib/setup-devices/setup-environment-ti
EdTheBearded: Awesome, thanks for the quick fix @lucas-akira!!!

jsrc27 (approved):

rborn-tx (approved):

PR #255, opened 2025-03-28
[Master] Maintenance

EdTheBearded:
Rebasing the commits from scarthgap-7.x.y that were missing on master branch, along with some fixes so that the build would complete.

The most notable change here would be commit: https://github.com/torizon/meta-toradex-torizon/commit/e8a1d0f3fd4d2014588d3204748a56377f56e308. Where masking `systemd-networkd-wait-online` was no longer working, so Systemd's `networkd` PACKAGECONFIG was disabled. Please check the commit message for more information.


jsrc27 (approved):

rborn-tx:
Btw, the solution of removing `networkd` from PACKAGECONFIG looks much better than masking the unit as we were doing before. Perhaps that should be cherry-picked to scarthgap too?

EdTheBearded:
> Btw, the solution of removing `networkd` from PACKAGECONFIG looks much better than masking the unit as we were doing before. Perhaps that should be cherry-picked to scarthgap too?

I think we could. Would even be a nice way to see if there'll be any side effects we're not anticipating as well (though I don't think so).
I'll open a PR cherry-picking this to Scarth and I'll mark as a draft while the release is underway!

rborn-tx:
@EdTheBearded I forgot to mention: there's typo on the commit message of the latest commit:

```
When we use systemd-networkd-wait-online, speciallu with --any flag, the
                                          ^^^^^^^^^
```

EdTheBearded:
> @EdTheBearded I forgot to mention: there's typo on the commit message of the latest commit:
> 
> ```
> When we use systemd-networkd-wait-online, speciallu with --any flag, the
>                                           ^^^^^^^^^
> ```

Nice!! I've fixed the comment both here and on Scarth PR.

PR #258, opened 2025-03-31
systemd: remove networkd PACKAGECONFIG from systemd

EdTheBearded:
On master branch, we're having a build issue where we can no longer mask the service 'systemd-networkd-wait-online.service', and this service is not behaving as it should.

When we use systemd-networkd-wait-online, speciallu with --any flag, the expected behavior is that it would wait for any one interface to become online to finish. But, since none of our interfaces are managed by systemd-networkd, it will wait until 2 minutes until it's internal timeout kicks in.
This ends up generating an impact, even though the device boots fine and is available, the serial interface will be blocked until the 2m timeout happens, and only then the login prompt will be shown, giving the idea that the device is taking longer to "boot".

So, since none of our interfaces are managed by systemd-networkd, 'networkd' PACKAGECONFIG was removed from systemd altogether.

Related-to: TOR-3512

(cherry picked from commit e8a1d0f3fd4d2014588d3204748a56377f56e308)


EdTheBearded:
Waiting for our quarterly releases to move it from draft.

jsrc27 (approved):
Just giving my preemptive approval.

rborn-tx (approved):
LGTM

PR #261, opened 2025-04-02
Fix Toradex TI builds

EdTheBearded:
Module required by TI's devices is not present on Toradex's BSP for TI, so we need to add that module only for the devices that require it.


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #265, opened 2025-04-10
u-boot-rollback: remove CONFIG_BOOTCOMMAND override in Torizon

EdTheBearded:
There's a legacy u-boot fragment that is overriding the 'bootcommand'. So far it wasn't causing any issues, since we were using a valid bootcommand. But with Aquila AM69, it is using 'bootflow' instead of distroboot, so this override was causing a boot issue.

Since this is something we get from Toradex's BSP, we can safely remove the bootcommand, but we leave the fragment there, if we need this for a device supported by Common Torizon.

Related-to: TOR-3800


rborn-tx (approved):
LGTM

PR #270, opened 2025-04-25
base-distro.inc: use default bootargs on all NXP machines

rborn-tx:
Machines apalis-imx6, colibri-imx6 and colibri-imx6ull were employing some additional boot arguments besides the ones normally used for booting Torizon OS. The additional arguments have been dropped by the BSP, so here we adjust the required bootargs to satisfy the kernel command line protection feature (relevant to the signed images).

Related-to: TOR-3828


jsrc27 (approved):
LGTM

PR #280, opened 2025-06-02
[Common] Fix iMX95 not booting

EdTheBearded:
Fixed iMX95 freezing during U-Boot initialization. Changed our recipe to stop masking `meta-imx`'s U-Boot recipe (where we used to use `meta-freescale`'s recipe).
This fixed our boot issues.

Also refactored our boot scripts, to use some more recent modifications we made.


jsrc27 (approved):
LGTM

rborn-tx (approved):
Nice!

PR #285, opened 2025-06-04
[Common] Use less disk space during builds

EdTheBearded:
Builds were using a lot of disk space, and was affecting negatively our CI.
On this change I've fixed the sstate-cache directory for TI builds, placing it on root directory instead of inside the build folder.

Also I've added a line that inherits `rm_work` bbclass for our builds, so that we cleanup a bit afterwards and use less disk space.


jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #289, opened 2025-06-11
Add new toradex devices to setup-environment scripts

EdTheBearded:
Add SMARC iMX8, iMX95 and Verdin AM62P.


rborn-tx (dismissed):
Raised a minor point but approving anyways.

PR #291, opened 2025-06-13
image_type_torizon: pack up files needed to resign OS images with TCB

lucas-akira:
As part of the effort to add in TorizonCore Builder the feature to resign existing signed Torizon OS images, TCB needs to recreate the bootloader container (imx-boot/flash.bin) from its constituent parts.

These files, composed of various U-Boot related binaries (SPL, U-Boot proper, U-Boot DTB, ATF, firmwares, etc.), need to be easily available in the image to be used by TCB. This TCB feature resigns the kernel fitImage as well, so it would also need to be made available.

This can be achieved by packing all required files in a tarball and including it when building the Easy Installer image. By default this will only be done when building images for supported machines that have TDX_IMX_HAB_ENABLE set to 1. This variable is automatically set when inheriting the 'tdx-signed' class from meta-toradex-security.

Users can also directly use TCB_SIGNING_SUPPORT to enable or disable the above behavior.

Currently the only supported machine is the Verdin iMX8M Plus. Other machines are planned to be supported in the future.

Related-to: TOR-3843


classes/image_type_torizon.bbclass
```diff
@@ -25,7 +25,48 @@ python adjust_tezi_artifacts() {
...
+                                      lpddr4_pmu_train_* u-boot.dtb u-boot-nodtb.bin spl/ dts/"
+
+pack_tcb_signing_binaries_in_teziimg() {
+    if [ "${TCB_SIGNING_SUPPORT}" != "1" ]; then
+        # TorizonCore Builder signing support feature not enabled
+        return 0
+    fi
+
+    if [ "${TDX_IMX_HAB_ENABLE}" != "1" ]; then
+        bbwarm "TCB signing support enabled but HAB support is disabled. Skipping."
```
EdTheBearded, line 41: shouldn't this be `bbwarn`? also on line 46

EdTheBearded:
Also, you said `resign`, but I figure you mean to sign again the image, no? This seems like you're doing a resignation (cancelling) of the build. Maybe reword to re-sign to avoid confusion?

classes/image_type_torizon.bbclass
lucas-akira: Oops, you're right. I'll correct it, thanks.

jsrc27 (dismissed):
Small bit of feedback, but overall looks okay to me.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -0,0 +1,13 @@
+# Deploy additional files needed by TorizonCore Buidler to recreate the bootloader container from its constituent parts
```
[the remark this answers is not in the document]
lucas-akira: Fixed, thanks!

EdTheBearded (approved):

jsrc27 (approved):

classes/image_type_torizon.bbclass
```diff
@@ -25,7 +25,48 @@ python adjust_tezi_artifacts() {
...
+TCB_SIGNING_FILELIST:verdin-imx8mp = "fitImage-initramfs-ostree-torizon-*.bin uboot_config bl31* \
+                                      lpddr4_pmu_train_* u-boot.dtb u-boot-nodtb.bin spl/ dts/"
+
+pack_tcb_signing_binaries_in_teziimg() {
+    if [ "${TCB_SIGNING_SUPPORT}" != "1" ]; then
+        # TorizonCore Builder signing support feature not enabled
+        return 0
+    fi
+
+    if [ "${TDX_IMX_HAB_ENABLE}" != "1" ]; then
```
rborn-tx, line 40: I think we also need to check `UBOOT_SIGN_ENABLE` here, no? Likewise for the default value of `TCB_SIGNING_SUPPORT`.

I mean, the code that follows requires that the kernel is in the form of a FIT image so if a user enables only bootloader (HAB) signing, an error would happen here when copying the FIT image file.

If we check if `UBOOT_SIGN_ENABLE == 1` I think it's enough in practice to infer that the `kernel-fitimage` class is in use and that the kernel is being signed which would be always the case in our `tdx-signed` builds.

classes/image_type_torizon.bbclass
```diff
@@ -25,7 +25,48 @@ python adjust_tezi_artifacts() {
...
+        return 0
+    fi
+
+    bbnote "Packing TorizonCore Builder signing files"
+
+    # Here we explicitly change to DEPLOY_DIR_IMAGE because apparently the shell resolves filenames
+    # with wildcards (e.g. with asterisk) relative to the directory the tar command was called, even
+    # when using the -C tar option
+    (cd "${DEPLOY_DIR_IMAGE}" && tar --preserve-permissions --dereference \
+        -czf "${WORKDIR}/${TCB_SIGNING_FILES_TARBALL}" ${TCB_SIGNING_FILELIST})
```
rborn-tx, line 56: For the record, the shell expands wildcards before invoking a command. E.g., assuming you have a file `fitImage-XZY.bin` inside `some/dir/`, the command:

```
tar -czf file.tgz -C some/dir/ fitImage*.bin
```

will be expanded to:

```
tar -czf file.tgz -C some/dir/ fitImage*.bin
```

i.e. **no expansion at all** because there is no file matching the wildcard `fitImage*.bin` in the current directory and the shell doesn't know that `tar` will grab the files from `some/dir/`.

If you did instead:

```
tar -czf file.tgz -C some/dir/ some/dir/fitImage*.bin
```

this would be expanded to:

```
tar -czf file.tgz -C some/dir/ some/dir/fitImageXYZ.bin
```

and the command would succeed, but, as a side effect, the files inside the tar would have the prefix `some/dir/` as part of their name which is not intended here. You could remove that suffix with the `--transform` switch but this would make this reliant on a GNU extension to `tar`.

So, entering the desired directory before calling `tar`, as you did, is the simplest solution.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -0,0 +1,13 @@
...
+
+do_deploy:append:verdin-imx8mp() {
+    install -m 0644 ${B}/${MACHINE}_defconfig/.config ${DEPLOYDIR}/uboot_config
+    install -d ${DEPLOYDIR}/spl
+    install -m 0644 ${B}/${MACHINE}_defconfig/spl/u-boot-spl ${DEPLOYDIR}/spl/
+    install -m 0644 ${B}/${MACHINE}_defconfig/spl/u-boot-spl.dtb ${DEPLOYDIR}/spl/
+    install -m 0644 ${B}/${MACHINE}_defconfig/spl/u-boot-spl-nodtb.bin ${DEPLOYDIR}/spl/
+    install -d ${DEPLOYDIR}/dts/upstream/src/arm64/freescale
+    install -m 0644 ${B}/${MACHINE}_defconfig/dts/upstream/src/arm64/freescale/imx8mp-verdin-wifi-dev.dtb \
+                    ${DEPLOYDIR}/dts/upstream/src/arm64/freescale/
```
rborn-tx, line 12: Initially I was thinking it would be better to use variable `UBOOT_DTB_NAME` here instead of hard-coding the DTB file name. However, I believe we don't have the information about the path in any OE variable either (since the location of the file is determined by U-Boot at build time by searching the file in multiple directories (IIRC)). So, doing this wouldn't fully solve the hard-coding part. I guess you considered this here, right?

Now, regarding the destination path for the file, is there any advantage of keeping the directory structure similar as in the source (i.e. `dts/upstream/src/arm64/freescale`)?

rborn-tx (changes requested):
Raised some questions and likely issue.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
lucas-akira: To be honest I didn't know about `UBOOT_DTB_NAME`. I did try searching for an OE/Yocto variable to get the DTB path in a more generic way, but I was unable to find one. So I ended up hard-coding it, but restricting its value for the compatible MACHINE.

>Now, regarding the destination path for the file, is there any advantage of keeping the directory structure similar as in the source (i.e. dts/upstream/src/arm64/freescale)?

I don't think so. I guess I kept it like that because the `binman` command execution from the build included the DTB with this structure.

I think the destination structure just needs to follow `CONFIG_OF_LIST` of the U-Boot config e.g. for the Verdin iMX8M Plus the value of `CONFIG_OF_LIST` is `freescale/imx8mp-verdin-wifi-dev`, so the DTB would just need to be inside a `freescale` directory.

I can change the destination path like above as that would make things simpler for the TCB implementation.

classes/image_type_torizon.bbclass
rborn-tx: Actually, the idea would be to use both: `TDX_IMX_HAB_ENABLE == "1"` means the bootloader is signed and `UBOOT_SIGN_ENABLE == "1"` means the kernel is a **signed FIT** image.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: It's up to you. I was asking just because having a particular directory structure felt like adding complexity. I mean, TCB would need to get the DTB name from the defconfig and traverse the tree to find the file somewhere or it would also hard-code a path which is probably worse.

classes/image_type_torizon.bbclass
```diff
@@ -25,7 +25,49 @@ python adjust_tezi_artifacts() {
...
+
+    cp "${WORKDIR}/${TCB_SIGNING_FILES_TARBALL}" \
+       "${IMGDEPLOYDIR}/${TCB_SIGNING_FILES_TARBALL}"
+}
+
+python add_signing_files_to_tezi_artifacts() {
+    signing_files_path = d.getVar('WORKDIR') + '/' + d.getVar('TCB_SIGNING_FILES_TARBALL')
+
+    if (os.path.isfile(signing_files_path)):
```
rborn-tx, line 66: Nitpick: the parentheses around the function call are not very Pythonic :wink:

recipes-bsp/u-boot/u-boot-tcb-sign.inc
lucas-akira: Sounds good to me. I'll follow your suggestion.

classes/image_type_torizon.bbclass
lucas-akira: Right, I'll add both checks in the function and in the default value for `TCB_SIGNING_SUPPORT`.

rborn-tx (approved):
Latest version looks good to me.

jsrc27 (approved):

PR #299, opened 2025-07-11
initramfs-module-kmod: add missing modules to enable native LVDS

EdTheBearded:
Related-to: TOR-3863


EdTheBearded:
> Is this for the release or for after?

Oh, right! This is for the release, this is a fix we'll need!!

PR #303, opened 2025-07-17
systemd: re-enable networkd PACKAGECONFIG from systemd

EdTheBearded:
networkd is needed to setup DHCP for the Access Point. As a faster solution to enable our release, we revert networkd, since this was an issue only on master branch, while we try to find a more permanent solution afterwards.

This reverts commit 4d0ebb190e6c15fd3f82c14aae4def2d4a6decec.

Related-to: TOR-3883


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #304, opened 2025-07-17
setup-environment-toradex: fix distro selection for imx95

EdTheBearded:
For the new devices (SMARC imx8mp/95 and Verdin am62p), imx8 and am62p distro selection were already covered, but imx95 was missing.

Now we add it to the rule so we select the downstream distro.

Related-to: TOR-3898


jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #310, opened 2025-07-22
[Kirkstone] sudo: upgrade 1.9.15p5 -> 1.9.17p1

EdTheBearded:
In order to fix sudo CVE's, we've ported the 'openembedded-cores' recipe
    to Torizon, while we internally bump our manifest.
    The sudo recipes were taken as is from openembedded-core kirkstone
    branch.

    Fixed CVE's:
    * Fixed CVE-2025-32462.  Sudo's -h (--host) option could be specified when running a command or editing a file.  This could enable a local privilege escalation attack if the sudoers file allows the user to run commands on a different host.

    * Fixed CVE-2025-32463.  An attacker can leverage sudo's -R (--chroot) option to run arbitrary commands as root, even if they are not listed in the sudoers file.  The chroot support has been deprecated an will be removed entirely in a future release.

Related-to: TOR-3913


jsrc27 (approved):
LGTM

PR #313, opened 2025-07-25
torizon.conf: Extending distro with formerly removed distro features

bstubert:
base-distro.inc removes features like 3g, alsa, nfc, and wayland from DISTRO_FEATURES. It makes impossible adding these features in a distro extending torizon.conf. Once an item is removed from a BitBake variable, it can never be appended again.

The solution is to assign the removed features to a new variable DISTRO_FEATURES_REMOVE in torizon.inc and :remove its contents from DISTRO_FEATURES in base-distro.inc.

Client code can remove features from DISTRO_FEATURES_REMOVE and append the features to DISTRO_FEATURES. DISTRO_FEATURES_REMOVE is an extension point for torizon.conf similar to IMAGE_FSTYPES_REMOVE. Now, clients can extend the distro conf torizon.conf.


EdTheBearded (approved):

jsrc27 (approved):
LGTM

PR #316, opened 2025-07-30
setup-environment-ti: make DL_DIR and SSTATE_DIR default values

EdTheBearded:
We're running with a problem on our automation scripts, where we wanted sstate-cache and downloads folder on a specific path. But we're using TI's setup environment script, and our automation scripts won't touch local.conf, only auto.conf. And local.conf has precedence over auto.conf.

So to in order to make things work and not make any considerable modifications, we can change the assignments of DL_DIR and SSTATE_DIR to default value assignments, which will only set a value to the variable if it isn't already initialized. This way we can work around the file precedence, and actually have our automation script values set on auto.conf take effect.


jsrc27 (approved):
LGTM

PR #319, opened 2025-08-05
toradex-net-rename: fix serial to MAC conversion

EdTheBearded:
Some devices had a serial number starting with a 0, which bash would interpret as a octal number and would generate the wrong MAC address for it.

Now we force every arithmetic to be base10 by using it with 'bc'.

Related-to: TOR-3849

### DIFF
--- a/dynamic-layers/meta-toradex-bsp-common/recipes-core/udev/files/toradex-net-rename.sh
+++ b/dynamic-layers/meta-toradex-bsp-common/recipes-core/udev/files/toradex-net-rename.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Assign name "ethernetN" to net devices with Toradex OUI.
 # Supports Toradex boards with single/dual ethernet interfaces.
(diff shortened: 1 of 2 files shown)


dynamic-layers/meta-toradex-bsp-common/recipes-core/udev/files/toradex-net-rename.sh
```diff
@@ -24,12 +24,12 @@ devaddr="${devpath}/address"
 case $oui in
   00:14:2d)
     max_range=16777216 # 16777215 = 0xFFFFFF, maximum valid serial for MAC OUI 1 (00:14:2D)
-    mac0_serial=$(( $(tr '\0' '\n' < /proc/device-tree/serial-number) + 0 ))
+    mac0_serial=$( echo "ibase=10; $(tr '\0' '\n' < /proc/device-tree/serial-number) + 0" | bc )
```
rborn-tx, line 27: @EdTheBearded You can set the base in Bash by prefixing the numeric string with the desired base. Some tests:

```
$ echo $(( 10#0100 - 100 ))
0

$ echo $(( $(echo 0100) - 100 ))
-36

$ echo $(( 10#$(echo 0100) - 100 ))
0
```

I guess this would still be better than resorting to an external program.

rborn-tx (commented):
Added a suggestion.

dynamic-layers/meta-toradex-bsp-common/recipes-core/udev/files/toradex-net-rename.sh
rborn-tx: Oh, I didn't realize the shebang was referring to just `sh` :weary:.

Well, IMO it would be fine to rely on `bash`. Looking at other scripts in `meta-toradex-torizon`, it seems today we don't have other scripts which run on the device that rely on `bc` while there are already a few depending on `bash`.

Anyways, it's not a strong opinion. If you prefer the `bc` approach it's fine as well. However, I think in that case we are missing an `RDEPENDS` on `bc` to be added into `udev-toradex-rules.bbappend`, no?

rborn-tx: Oh, in case we go with Bash, we may need the `RDEPENDS` for it as well unless `udev` already defines that dependency.

rborn-tx (approved):
LGTM

PR #320, opened 2025-08-05
[Master] toradex-net-rename: fix serial to MAC conversion

EdTheBearded:
Some devices had a serial number starting with a 0, which bash would interpret as a octal number and would generate the wrong MAC address for it.

Now we force every arithmetic to be base10 by using it with 'bc'.

Related-to: TOR-3849

(cherry picked from commit f2309c7a14c105e821f35ed5eb7fd39d97eff2b2)


rborn-tx (approved):
LGTM

PR #329, opened 2025-08-18
[Master] initramfs-framework: refactor and add missing module for SMARC iMX8MP

EdTheBearded:
Removed leftover modules on for TI devices, refactored the way we add modules to minimize duplication and file creation and added missing module to have splash working on SMARC iMX8MP.

Related-to: TOR-3894


rborn-tx (approved):
LGTM

PR #330, opened 2025-08-25
scope: OEROOT path is broken when building out of tree

cgfacts:
Attempting to build out of tree, for example with the command ```MACHINE=verdin-imx8mp EULA=1 source setup-environment ~/build-torizon-docker/```, fails due to how OEROOT is being set in the bblayers.conf template file.  See [yocto documentation](https://docs.yoctoproject.org/ref-manual/structure.html#build)

Update bblayers.conf template file and setup-environment-toradex to allow for out of tree build directory.

### DIFF
--- a/conf/template/bblayers.conf
+++ b/conf/template/bblayers.conf
@@ -1,7 +1,7 @@
 # LAYER_CONF_VERSION is increased each time build/conf/bblayers.conf
 # changes incompatibly
 LCONF_VERSION = "7"
-OEROOT := "${@os.path.abspath(os.path.dirname(d.getVar('FILE', True)))}/../.."
+OEROOT := "@OEROOT@"
 
 BBPATH = "${TOPDIR}"
 
--- a/scripts/lib/setup-devices/setup-environment-toradex
+++ b/scripts/lib/setup-devices/setup-environment-toradex
@@ -103,7 +103,8 @@ sha512sum "${_TDX_SCRIPTS}"/lib/setup-devices/setup-environment-toradex 2>&1 > c
 if [ ! -f "conf/local.conf" ]; then
     cp "${_TDX_SCRIPTS}"/../conf/template/local.conf conf/local.conf
 fi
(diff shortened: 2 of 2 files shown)


rborn-tx:
I see that what's being proposed here is pretty much the same that is done on Poky:

https://github.com/yoctoproject/poky/blob/4ad510a63fd88b7907b2a1b8feceace9758397d0/scripts/oe-setup-builddir#L92

1. It makes a copy of the `bblayers.conf` file from the source directory rather than making a symlink.
2. It hard-codes the value of the OEROOT into that copy (by replacing some text into that copy, thus the need of a copy).

Being more aligned with Poky looks sensible to me but notice this comes with a side-effect: if the `bblayers.conf` file is updated in the source directory (e.g. by a `repo sync`), it won't be updated in the build directory as the file is now a copy. This is probably the main point to be considered here before accepting this contribution.

Personally I'd be okay with the change in behavior, but I'm not sure if this could impact reasonable use cases.

@jsrc27 @EdTheBearded Opinions?

Having said that, I'd say that the way we set up the directory structure already follows the good practice of separating source (`metatada = layers`) and build files. Notice below anything produced during a particular build is inside that build directory while the source comes from the `layer/` directory one level above:

```
.
├── build
│   ├── conf
│   ├── deploy   <== final output
│   └── tmp      <== temporary output
│       ├── work
│       └── work-shared
├── downloads
├── layers
│   ├── manifests
│   ├── meta-arm
│   ├── ...
│   ├── meta-toradex-security
│   ├── meta-toradex-ti
│   ├── meta-toradex-torizon
│   ├── meta-yocto
│   └── openembedded-core
└── sstate-cache
```

What is not currently allowed is placing the `build` directory on an arbitrary location, which it seems would be possible if we followed the Poky approach.

EdTheBearded:
Thanks @rborn-tx!

I think it's fine to have this different behavior. And would actually be nice to stop depending on relative paths, when we separate our outputs, as you pointed out.

cgfacts:
Thank you for approving.  I updated the commit message as suggested.

rborn-tx (approved):

PR #331, opened 2025-08-25
setup-environment-ti: add a retry on downloading content from TI

EdTheBearded:
Connection with 'git.ti.com' fails consistently, so we add a retry option when we donwload content from there.


jsrc27 (approved):
LGTM

PR #333, opened 2025-09-03
images/torizon-base.inc: Add rogue drivers for AM62P target

flavioaraujodex:
This add support to AM62P hardware in the same way we had for AM62 target. Without this we don't have accelerated features while using Chromium container.

Related-to: TCCP-1046

### DIFF
--- a/recipes-images/images/torizon-base.inc
+++ b/recipes-images/images/torizon-base.inc
@@ -94,6 +94,10 @@ CORE_IMAGE_BASE_INSTALL:append:am62xx = " \
     ti-img-rogue-driver \
 "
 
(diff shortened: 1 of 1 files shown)


EdTheBearded:
Hey @flavioaraujodex, is this something we want on both Verdin AM62P and TI AM62P (Common Torizon)?
If not, I'd suggest we use another override, such as `verdin-am62p`.

EdTheBearded (approved):
LGTM.
Let's just wait for the monthly to finish today, then we can merge this.

EdTheBearded:
@flavioaraujodex could you please cherry-pick this PR to the master branch as well?

PR #335, opened 2025-09-05
rpcbind: Disable systemd service

jsrc27:
There are some minor security concerns with having the rpcbind service enabled by default. This is because the service listens on 0.0.0.0.

The risk is fairly minor, but it's better to disable this as the default behavior. Users can re-enable this on a per use-case basis.

Related-to: TOR-3959


rborn-tx (approved):
LGTM

EdTheBearded (approved):

EdTheBearded:
@jsrc27  could you please cherry-pick this PR to the master branch as well?

PR #336, opened 2025-09-10
images/torizon-base.inc: Add rogue drivers for AM69 target

flavioaraujodex:
This add support to AM69 hardware in the same way we had for others AM6* targets. Without this we don't have accelerated features while using Chromium container.

Related-to: TCCP-1058


EdTheBearded:
@flavioaraujodex could you please cherry-pick this PR to the master branch as well?

PR #344, opened 2025-09-17
initramfs-graphics: Add missing modules for Aquila AM69 splash screen

jsrc27:
Related-to: TOR-3721

### DIFF
--- a/recipes-core/images/initramfs-ostree-torizon-image.bb
+++ b/recipes-core/images/initramfs-ostree-torizon-image.bb
@@ -9,6 +9,11 @@ PACKAGE_INSTALL:append:cfs-support = "\
     initramfs-module-composefs \
 "
 
+# Additional firmware needed for splash screen on DisplayPort with Aquila AM69
+PACKAGE_INSTALL:append:aquila-am69 = "\
+    cadence-mhdp-fw \
+"
+
 SYSTEMD_DEFAULT_TARGET = "initrd.target"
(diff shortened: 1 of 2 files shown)


jsrc27:
One small note. As noted in the comment for the `i2c_mux_pca954x` kernel module. This is only needed for V1.3 of the Aquila dev carrier board. I only have V1.2 of the carrier board (which works fine without this module), so I couldn't explicitly test this. That said, I can see in the device tree for V1.3 that this is needed, and it's used in the same way as with the SMARC IMX8MP. Therefore I'm fairly confident it should be added to the initramfs here as well. So I decided to just preemptively add this module despite not having the V1.3 yet.

EdTheBearded:
> One small note. As noted in the comment for the `i2c_mux_pca954x` kernel module. This is only needed for V1.3 of the Aquila dev carrier board. I only have V1.2 of the carrier board (which works fine without this module), so I couldn't explicitly test this. That said, I can see in the device tree for V1.3 that this is needed, and it's used in the same way as with the SMARC IMX8MP. Therefore I'm fairly confident it should be added to the initramfs here as well. So I decided to just preemptively add this module despite not having the V1.3 yet.

And what about v1.2 and lower? They use another driver or they just don't support it? Also, I assume no but just to clarify, it won't affect the build, right?

rborn-tx (approved):
LGTM

PR #350, opened 2025-10-06
linux-firmware: remove unnecessay firmware files

EdTheBearded:
For some reason, having the firmwares from 'rtl_bt' installed, was causing for every firmware to be installed in the final image, which resulted in a image size growth of ~70% and some compilation errors for Colibri and Apalis iMX8 (devices unrelated to the original change).

Since 'rtl_bt' firmware are not required, we simply skip their installation.

Related-to: TOR-4027


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #446, opened 2026-01-29
aktualizr-default-sec: fix QA issue files not being shipped

EdTheBearded:
After the modifications from 2ee819b2504797c44612bb9217193e51e2a381c2, installation of scripts were conditional, but the creation of `bindir` was not, which was failing QA test.

So to fix this, we create the `bindir` only when we're actually going to install binaries to it.


rborn-tx (approved):
LGTM

jsrc27 (approved):

MatheusRodrigues-tor (approved):

PR #452, opened 2026-01-30
intel-corei7-64: Add RTL8XXXU kernel modules for USB WLAN.

drewmoseley:


drewmoseley:
Not quite ready for merge.  The module is there but does not load automatically.  And even if I manually modprobe it, the interface does not show up.  Maybe I need more firmware?

drewmoseley:
OK. This is now functional and ready for review

drewmoseley:
@EdTheBearded @jsrc27 @rborn-tx can you guys expedite this review? I have a user wanting to test it in a nightly.

dynamic-layers/meta-intel/recipes-kernel/linux/linux-intel%.bbappend
```diff
@@ -9,3 +9,5 @@ DEPENDS += "lz4-native"
 SRC_URI += "\
     file://intel-corei7-64.cfg \
 "
+
+RDEPENDS_${PN}:append = " kernel-module-rtl8xxxu linux-firmware-rtl8188 "
```
EdTheBearded: You're already enabling RTL8XXXU as a module on the fragment, why do we need the kernel module here?

drewmoseley: Enabling the module in the config ensures that it is built and packaged as the kernel-module-rtl8xxxu package but that does not necessarily mean it is included in the image right?  Or do we somehow automatically include any kernel-module-* packages that are built? I don't believe in poky they do that but I may be unaware of the mechanisms here.

EdTheBearded: Right, but in our image we're installing the `kernel-modules` package:
https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/recipes-images/images/torizon-base.inc#L40

This is a package that will install every module built into the final image.

drewmoseley: OK.  thanks for clarifying.  In that case I guess we don't need it.  I'll remove it, rebuild and test.

drewmoseley: @EdTheBearded updated, built and tested. This is ready for re-review.

jsrc27 (approved):

EdTheBearded (approved):

rborn-tx (approved):
LGTM

PR #456, opened 2026-02-10
base-distro: use weak assignment for USERADD variables

hiagofranco:
Use weak assignment for USERADD* variables. In case there is a need to modify the passwd and group files of Torizon, this makes it easier to change USERADD values with different layers.


jsrc27 (approved):
LGTM

EdTheBearded (approved):

rborn-tx (approved):
LGTM

PR #458, opened 2026-02-13
README-nvidia: Fix recovery button header and add link to manual

LeoBelmont:


docs/README-nvidia.md
```diff
@@ -83,7 +83,7 @@ echo "Removing temp directory $tmpdir"
 rm -rf $tmpdir
 ```
 
-3. Connect the USB port to the host machine and put the target board into the recovery mode: short the `FORCE_RECOVERY` pin (labeled as `FC_REC` on the carrier board) to the ground, this can be done by installing a wired jumper to the contacts `9-10` of the `J50` `button` header on the carrier board (refer to the Jetson Orin Nano Developer Kit Carrier Board Specification), then power on the board.
+3. Connect the USB port to the host machine and put the target board into the recovery mode: short the `FORCE_RECOVERY` pin (labeled as `FC_REC` on the carrier board) to the ground, this can be done by installing a wired jumper to the contacts `9-10` of the `J14` `button` header on the carrier board (refer to the [Jetson Orin Nano Developer Kit Carrier Board Specification](https://developer.nvidia.com/downloads/assets/embedded/secure/jetson/orin_nano/docs/jetson_orin_nano_devkit_carrier_board_specification_sp.pdf)), then power on the board.
```
EdTheBearded, line 86: @LeoBelmont this is the link with the info on how to partition the NVMe?
Do you think we could bring some of those steps here?

LeoBelmont: This is the NVIDIA board specification showing the pins location. I will also add the steps for partitioning.

rborn-tx (dismissed):
LGTM

jsrc27 (dismissed):

docs/README-nvidia.md
LeoBelmont: @rborn-tx @jsrc27 @EdTheBearded reopening this thread to check the partitioning steps again. @griloHBG was helping me test and it didn't work on the SD card, we need to verify if some step was missed or something else went wrong. Due to holidays it may take a few days to retest

MatheusRodrigues-tor (approved):

EdTheBearded:
@LeoBelmont in light of the latest discoveries we did on the flashing process, would you agree to move this to draft while we sort things out?
My main point here would be to try and avoid merging something now that we would need to change a bit later. What do you think?

LeoBelmont:
@EdTheBearded @rborn-tx @MatheusRodrigues-tor @jsrc27 I removed the commit about the partitioning steps as we found out that they are not needed and the issue only happens on some environments. We are not sure why the issue happens yet.

Because of this it is better to drop the commit and only merge what we are 100% sure of (The wrong pins and link to manual).

I also removed the draft status from the PR.

rborn-tx (approved):

jsrc27 (approved):

EdTheBearded (approved):

PR #470, opened 2026-03-17
initramfs-framework: add lost patch mounting /run

EdTheBearded:
Up until scarthgap (and master) we used to have this patch[1], which is necessary for ostree to run ostree-system-generator. At some point, this change made it to upstream[3], and so we had to drop it from our master branch[2].

When we forked our master to create the walnascar branch, we ended up with half measures: we had the patch removed on our end and it's also not available in walnascar branch of openembedded-core. This is why we're seeing the issue where no home directory exists.

And to fix this, a backport of OE-core main branch[3] was performed,

Note: Due to the fact that, historically this patch was the first one applied[1], and that our patch names were kept '0002' and '0003' (no '0001'), I've decided to add this back as '0001', and with this I needed to change '0003-notify-newroot-for-plymouth.patch', since the context changed.

[1] https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/recipes-core/initramfs-framework/files/0001-Mount-run-with-tmpfs.patch
[2] https://github.com/torizon/meta-toradex-torizon/commit/02a7ae134f5f60e76f5fd05b3d513bb7c1171c64
[3] https://github.com/openembedded/openembedded-core/commit/3a4bd7ddefbf5b412a2b4031d491f5a50f1908cd

Related-to: TOR-4238

### DIFF
--- a/recipes-core/initramfs-framework/files/0001-initramfs-framework-mount-run-and-move-to-rootfs-bef.patch
+++ b/recipes-core/initramfs-framework/files/0001-initramfs-framework-mount-run-and-move-to-rootfs-bef.patch
@@ -1,0 +1,64 @@
+From b5d8f0f872037e437a0f55037e422fb37501f420 Mon Sep 17 00:00:00 2001
+From: Eduardo Ferreira <eduardo.barbosa@toradex.com>
+Date: Tue, 17 Mar 2026 13:13:15 +0000
+Subject: [PATCH] initramfs-framework: mount /run and move to rootfs before
+ switch_root
+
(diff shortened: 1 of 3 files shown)


rborn-tx (approved):
LGTM

Freireg (approved):
LGTM

MatheusRodrigues-tor (approved):
LGTM

EdTheBearded:
@rborn-tx, can we merge this? I believe @jsrc27 is on vacation :grinning:

PR #478, opened 2026-03-25
ostree: allow pulling local repos with static-deltas

rborn-tx:
Add a patch to libostree to change its default behavior when pulling commits from local repositories: instead of disabling static-deltas, allow them to be used if present.

The immediate goal is to allow Aktualizr (user of libostree) to load Lockboxes having these static-deltas.

Related-to: TOR-4248


jsrc27 (approved):
LGTM

EdTheBearded (approved):
LGTM
Could you also cherry-pick this to master once it's done?

PR #503, opened 2026-04-28
Generate WIC image when building with class tdx-signed

rborn-tx:
Enable the generation of a WIC image when building Torizon images with class `tdx-signed`; this covers also child classes such as `torizon-signed`.

The resize-helper service responsible for resizing the main partition to occupy the available space on disk is also installed and executed, but only on `tdx-signed`. On `torizon-signed` the partition resizing is also required but tests showed adaptations would be needed to make it work when the rootfs protection is in place.

Related-to: TOR-4364
Signed-off-by: Rogerio Guerra Borin <rogerio.borin@toradex.com>


jsrc27 (approved):
LGTM

Freireg (approved):

PR #504, opened 2026-05-04
Add flashing files for Luna SL1680

jsrc27:
This makes it so that the flashing files specific for the Luna SL1680 get packaged when building for this machine.

### DIFF
--- a/recipes-support/syna-usb-tool/syna-usb-tool_git.bb
+++ b/recipes-support/syna-usb-tool/syna-usb-tool_git.bb
@@ -8,6 +8,11 @@ PV = "1.0.5"
 SRC_URI = "git://github.com/synaptics-astra/usb-tool.git;protocol=https;branch=main"
 SRCREV = "e5e9a160ffb1755549e73457bda767e641439dc6"
 
+SRC_URI:append:luna-sl1680 = " \
+    https://artifacts.toradex.com/artifactory/common-torizon-extras-prod-frankfurt/luna-sl1680/astra-usbboot-images-winglet.tar.gz;name=luna-tarball \
+    "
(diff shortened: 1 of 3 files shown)


MatheusRodrigues-tor (dismissed):
LGTM

recipes-support/syna-usb-tool/syna-usb-tool_git.bb
```diff
@@ -23,6 +28,11 @@ do_deploy() {
     cp -r ${S}/astra-usbboot-images/ ${SYNAIMG_DEPLOY}
 }
 
+# Replace with Luna-specific files from tarball
+do_deploy:append:luna-sl1680() {
+    cp -r ${WORKDIR}/astra-usbboot-images/ ${SYNAIMG_DEPLOY}
```
rborn-tx, line 33: Just to confirm: shouldn't the `-p` switch be used here to ensure ownership and permissions are kept (especially if there are executable files/scripts in the tarball)?

rborn-tx (commented):
Looks go to me overall. Raised a question though...

recipes-support/syna-usb-tool/syna-usb-tool_git.bb
jsrc27: From my own tests, this did not seem required. Also, none of the files provided in the tarball are executables used for the image flashing.

Similar to the `astra-usbboot-images` that comes from the Git repo in the above `do_deploy`. These files need to just be present in the final SYNAIMG deployment.

jsrc27:
Just pushed a small update to the PR as I noticed some comments needed to be updated in the `flash-image.sh` script.

Freireg (approved):

recipes-support/syna-usb-tool/syna-usb-tool_git.bb
rborn-tx: Got it. Things look fine then.

rborn-tx (approved):

PR #512, opened 2026-05-07
Fix master build

jsrc27:
Some fixes so that Torizon can build successfully on master. See individual commits for more details.


jsrc27:
Small side-note, none of this was runtime-tested. The only test was ensuring the Yocto build could conclude successfully. This will help us as we begin to do some initial work and testing on master.

We can probably worry about runtime later, as getting the build to complete is more important.

Freireg (approved):

EdTheBearded (approved):
Thanks for the work maintaining master! 😉

MatheusRodrigues-tor (approved):

rborn-tx (approved):
LGTM

PR #515, opened 2026-05-07
plymouth.conf: Increase Plymouth's default DeviceTimeout to 15 seconds.

Freireg:
This is a workaround for the intermittent failure during the Splash Screen on the aquila-imx95 boot. To avoid possible race conditions during boot, the default timeout was increased from 8 to 15 seconds. 
The additional seconds seems to grant enough time for the plymouth-quit service to run as intended without compromising the boot time too much (adding as much as half a second to "bad" boots).

(This solution also works for smarc-imx95 machines)

Related-to: TOR-4307

### DIFF
--- a/recipes-core/plymouth/files/plymouthd.conf
+++ b/recipes-core/plymouth/files/plymouthd.conf
@@ -1,0 +1,2 @@
+[Daemon]
+DeviceTimeout=15
--- a/recipes-core/plymouth/plymouth_%.bbappend
+++ b/recipes-core/plymouth/plymouth_%.bbappend
@@ -4,6 +4,7 @@ SRC_URI += " \
     file://0001-disable-boot-splash-later.patch \
     file://torizonlogo-white.png \
     file://spinner.plymouth \
+    file://plymouthd.conf \
 "
 
 PLYMOUTH_THEMES = "spinner"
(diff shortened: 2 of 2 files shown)


jsrc27:
Question. Does this have an effect on boot times for all devices?

If so, we may need to discuss this further. Some users can be very sensitive about boot times, even with increases of half a second.

Freireg:
@jsrc27 the timeout only adds time to boot if the timing issue actually occurs. On devices where the issue is not present, the higher ceiling should have no effect. Also, the half-second figure was the worst case observed during testing. 

Considering this, do you think it makes sense to run some benchmarks with other machines to assure that they aren't being impacted? Let me know what you think

jsrc27:
> Considering this, do you think it makes sense to run some benchmarks with other machines to assure that they aren't being impacted? Let me know what you think

As long as we're confident it has no detrimental effects on other machines, where the root issue isn't present. That was my main concern.

jsrc27 (approved):

rborn-tx (approved):
LGTM (assuming that it does not negatively impact setups where the issue is not present)

MatheusRodrigues-tor (approved):

EdTheBearded (approved):

EdTheBearded:
@Freireg please, cherry-pick this MR to master

PR #536, opened 2026-06-09
[master] torizon-base: Remove GPLv3 dosfstools

Freireg:
Removed dosfstools package.
Related-to: TOR-4380

PR #539, opened 2026-06-11
[master] gplv3-package-filter: create GPLv3 filtering class for the distro

Freireg:
The gplv3-package-filter.bbclass will parse each recipe used by the distro and strip any `RRDEPENDS` defined in `TORIZON_EXCLUDED_PACKAGES_GPL_V3`, this enables a centralized list to drop any dependency related to the GPLv3 removal. Also removed the `BAD_RECOMMENDATIONS_GPL_V3` in favor of using the same variable.

Removed `packagegroup-core-full-cmdline.bbappend` since it's no longer needed in this setup.

Related-to: TOR-4381

### DIFF
--- a/classes/gplv3-package-filter.bbclass
+++ b/classes/gplv3-package-filter.bbclass
@@ -1,0 +1,35 @@
+# Remove the packages listed in TORIZON_EXCLUDED_PACKAGES_GPL_V3 from the
+# RDEPENDS/RRECOMMENDS of every package of every recipe, so a GPLv3 package
+# can be dropped from the build by listing it in a single place instead of
+# one .bbappend per recipe that pulls it in.
+#
+# This class is inherited globally from base-distro.inc and the package list
+# is defined there as well.
+#
+# Note: this covers runtime dependencies declared in recipe metadata (e.g.
+# packagegroups). Dependencies generated during do_package (shared library
+# auto-dependencies, dynamic packages) are not affected; BAD_RECOMMENDATIONS
+# in base-distro.inc act as the image-level backstop for those.
+
+TORIZON_EXCLUDED_PACKAGES_GPL_V3 ??= ""
+
+python () {
+    excluded = set((d.getVar('TORIZON_EXCLUDED_PACKAGES_GPL_V3') or '').split())
+    if not excluded:
+        return
+
+    # Host tools may be GPLv3; only filter target recipes.
+    for native_class in ('native', 'nativesdk', 'cross', 'crosssdk'):
+        if bb.data.inherits_class(native_class, d):
+            return
+
+    for pkg in (d.getVar('PACKAGES') or '').split():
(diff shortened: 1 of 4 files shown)


Freireg:
This implementation already covers the recent MRs https://github.com/torizon/meta-toradex-torizon/pull/532, https://github.com/torizon/meta-toradex-torizon/pull/536 and https://github.com/torizon/meta-toradex-torizon/pull/537

jsrc27:
Wouldn't this be "too global" of an affect?

For example, if a user is doing a custom Torizon OS Yocto build. They might not care about this whole GPLv3 stuff. So they add a package they want to their custom build that `RDEPENDS` on one of the listed GPLv3 packages. Maybe this new package really needs this `RDEPENDS` so their build breaks. It wouldn't be immediately obvious to them why it broke and what they need to do to fix it.

Furthermore, even if they do fix it by knowing to modify this new distro level variable, there's no way to add the `RDEPENDS` to just their new package, correct? If they remove something from that variable, it would add that `RDEPENDS` back in everywhere. Which makes it harder to do precise controlled changes in their builds.

Thoughts everyone?

classes/gplv3-package-filter.bbclass
```diff
@@ -0,0 +1,35 @@
...
+
+    for pkg in (d.getVar('PACKAGES') or '').split():
+        for depvar in ('RDEPENDS:' + pkg, 'RRECOMMENDS:' + pkg):
+            value = d.getVar(depvar)
+            if not value:
+                continue
+            deps = bb.utils.explode_dep_versions2(value)
+            filtered = {name: constraint for name, constraint in deps.items() if name not in excluded}
+            if len(filtered) != len(deps):
+                d.setVar(depvar, bb.utils.join_deps(filtered, commasep=False))
```
rborn-tx, line 34: Can you explain what this is doing and why it is needed?

rborn-tx (changes requested):
Typo in commit message and PR description: "classe" -> "class"

Blindly removing packages from `RDEPENDS` seems very dangerous to me and should probably be discussed as we'd be running the risk of removing a dependency to something that is really needed.

Worse: the build would likely pass (because the removed dependencies are runtime ones) while failures would be seen only at runtime. But our tests might not exercise the cases to catch those failures.

classes/gplv3-package-filter.bbclass
Freireg: The `explode_dep_versions2(value)`  is bitbake's own parser for dependency strings, it returns a dict mapping each package name to its version constraints (I think I could've used `explode_dep_versions()`, since it does seem to accomplish the same).
The one-liner in `filtered` selects only the packages that are not excluded and create a new dict. 
And after a final validation that anything has really changed, we undo the `explode_dep_versions2(value)` by serializing the dict and writing the cleaned string back into the recipe's datastore under the same variable name it came from.

PR #548, opened 2026-06-16
apparmor: Add apparmor userspace utils

jsrc27:
Adds packages to enable the use of apparmor and it's userspace utilities.

Related-to: TOR-4439

### DIFF
--- a/recipes-images/images/torizon-base.inc
+++ b/recipes-images/images/torizon-base.inc
@@ -25,6 +25,7 @@ SPLASH = "plymouth"
 
 # Base packages
 CORE_IMAGE_BASE_INSTALL:append = " \
+    apparmor \
     cpupower \
     curl \
     ethtool \
--- a/conf/distro/include/base-distro.inc
+++ b/conf/distro/include/base-distro.inc
@@ -40,7 +40,7 @@ IMAGE_BOOT_FILES = ""
 IMAGE_BOOT_FILES_REMOVE:sota = "boot.scr-${MACHINE};boot.scr overlays.txt overlays/*;overlays/"
 IMAGE_BOOT_FILES_REMOVE:append:apalis-imx8 = " hdmitxfw.bin dpfw.bin"
 
-DISTRO_FEATURES:append = " virtualization stateless-system"
+DISTRO_FEATURES:append = " virtualization stateless-system apparmor"
 DISTRO_FEATURES_REMOVE ?= "3g alsa irda pcmcia nfc ldconfig pulseaudio wayland x11 ptest multiarch vulkan"
 DISTRO_FEATURES_REMOVE:remove:mx95-nxp-bsp = "vulkan"
 DISTRO_FEATURES_REMOVE:append:imx-generic-bsp = " opengl"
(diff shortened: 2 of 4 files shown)


recipes-images/images/torizon-base.inc
```diff
@@ -50,6 +51,7 @@ CORE_IMAGE_BASE_INSTALL:append = " \
     packagegroup-core-full-cmdline-multiuser \
     packagegroup-core-full-cmdline-utils \
     parted \
+    python3-cgitb \
```
rborn-tx: Is this `python3-cgitb` package being added to support `apparmor`? If so, shouldn't `apparmor` itself pull that dependency in?

jsrc27: Several of the apparmor userspace utilities rely on `python3-cgitb` in order to actually work properly without error.  For whatever reason, the `apparmor` recipe: https://git.yoctoproject.org/meta-security/tree/recipes-mac/AppArmor/apparmor_3.1.3.bb?h=scarthgap

Does not pull this in. I agree with you it should but it doesn't, not even on `master`.

EdTheBearded: This is so weird....
But on master, they have this patch on the recipe, to make the include optional:
https://git.yoctoproject.org/meta-security/tree/recipes-mac/AppArmor/files/0001-fail.py-handle-missing-cgitb.patch?h=master

Maybe we can backport this patch as well?

jsrc27: > Maybe we can backport this patch as well?

Hmm what's preferable for us from a maintenance point of view? 

Either way we carry something extra in Scarthgap, a patch versus a package.

EdTheBearded: I'd rather a patch, since it's on master already.

rborn-tx: I too think the patch is preferable.

Though we may need another patch, as per commit `5c98ff10a65534dffd2e6426ea8f5592f3957a2d` message where that  patch was added:

```
apparmor: update from 3.1.3 to 4.0.3

Fixes python 3.13 support though needed one more patch
which is also submitted upstream. oeqa runtime test
passes on qemuarm and qemuarm64. Did not fix ptest compilation.
```

jsrc27: Patch backported. I only needed the one patch it seems. The build succeeds, and none of the apparmor utils show any python related errors anymore during runtime.

EdTheBearded: \o/

EdTheBearded:
And just to confirm, this is something we're enabling for all images right? Not something that should be restricted to signed images?

jsrc27:
> And just to confirm, this is something we're enabling for all images right? Not something that should be restricted to signed images?

Correct. Anyways, `apparmor` is runtime disabled by default. It requires the user to consciously set `security=apparmor` on the kernel command-line to fully enable it.

The idea is to give users the flexibility, and choice if they're interested in `apparmor` without affecting the majority of users who likely don't care about this feature.

recipes-images/images/torizon-base.inc
rborn-tx: Awesome. Thanks for making the changes.

rborn-tx (approved):

Freireg (approved):

EdTheBearded (approved):

MatheusRodrigues-tor (approved):

PR #550, opened 2026-06-18
[master] libbytesize: remove GPLv3 mpfr

Freireg:
Remove the dependency from `libbytesize` recipe.
Requires the backported patches to remove the hardcoded dependecy on the `mpfr` C library. These are already on upstream, and should be dropped as soon as a new release is out.

Related-to: TOR-4389

### DIFF
--- a/recipes-support/libbytesize/libbytesize_%.bbappend
+++ b/recipes-support/libbytesize/libbytesize_%.bbappend
@@ -1,0 +1,12 @@
+# Remove mpfr as a dependency due to GPLv3.
+# Requires the backported patches to remove the hardcoded dependecy on the
+# mpfr C library
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI:append = " \
+   file://0001-Fix-issues-with-floating-point-precision-and-roundin.patch \
+   file://0002-Add-regression-tests-for-parsing-edge-cases.patch \
+   file://0003-Fix-number-parsing-edge-cases.patch \
+"
+
+DEPENDS:remove = "mpfr"
--- a/recipes-support/libbytesize/files/0001-Fix-issues-with-floating-point-precision-and-roundin.patch
+++ b/recipes-support/libbytesize/files/0001-Fix-issues-with-floating-point-precision-and-roundin.patch
@@ -1,0 +1,482 @@
+From 4797c5d66c4541c5727958960f638290744f58f2 Mon Sep 17 00:00:00 2001
+From: Jon Oster <jon.i.oster@gmail.com>
+Date: Wed, 17 Dec 2025 15:56:47 -0800
+Subject: [PATCH] Fix issues with floating point precision and rounding
+
+This commit started off because I wanted to remove MPFR as a dependency.
+After an initially-incorrect attempt at doing so, I discovered two more
+problems. First, the test suite didn't have any tests verifying that MPFR
+actually fixed the problem it was introduced to fix (correctly rounding
+results like 0.1*1000 to 100 instead of 99). Second, MPFR gets 0.1*1000,
+but still makes rounding mistakes in other cases--I found this out by
+fuzzing after my initial incorrect patch.
+
+So now I've replaced MPFR with exact rational math in libgmp, and added
+regression tests for the rounding errors in string parsing. To do this, I
+had to do the parsing manually; I followed the same regex approach that
+the existing function uses.
+
+Signed-off-by: Jon Oster <jon.i.oster@gmail.com>
+
+Upstream-Status: Backport [https://github.com/storaged-project/libbytesize/commit/4797c5d66c4541c5727958960f638290744f58f2]
+---
+ configure.ac                  |   2 +-
+ po/libbytesize.pot            |   2 +-
+ src/Makefile.am               |   2 +-
+ src/bs_size.c                 | 176 ++++++++++++++++++++++++----------
+ src/bytesize.pc.in            |   2 +-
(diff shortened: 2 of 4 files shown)


recipes-support/libbytesize/libbytesize_%.bbappend
```diff
@@ -0,0 +1,12 @@
...
+# Patches not cleaning applying on older version, let's just bump versions for now.
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI:append = " \
+   file://0001-Fix-issues-with-floating-point-precision-and-roundin.patch \
+   file://0002-Add-regression-tests-for-parsing-edge-cases.patch \
+   file://0003-Fix-number-parsing-edge-cases.patch \
+"
+
+SRCREV = "518fd4a0a863e896675a7f5ef0164a36eb4d7b03"
```
rborn-tx: Forcing a `SRCREV` doesn't seem like a good thing for us to do here (in an append that applies to any version of `libbytesize`) because we'll not follow the upstream versions as they evolve. Plus we would silently diverge from upstream, as, btw, is already happening: this particular SRCREV is for version 2.12 whereas the recipe in upstream OpenEmbedded is version 2.11.

I think it would be better to rebase and adapt the patches on top of 2.11. Then when the upstream is updated, our build would likely break and we'd be reminded to drop (or rebase again) the patches.

@EdTheBearded @jsrc27 Opinions?

rborn-tx:
Nitpick: typo in commit message: `dependecy`

recipes-support/libbytesize/libbytesize_%.bbappend
jsrc27: Wait isn't master/upstream for this recipe on 2.12 already?: https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/libbytesize/libbytesize_2.12.bb?h=master

Which makes this `SRCREV` redundant. Or, am I misunderstanding something?

rborn-tx: The version pinned in our manifest is still on 2.11: https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/libbytesize/?h=b22026cd3785700052aefa90e709b92d8e7f74fc

Freireg: You're right @rborn-tx the `SRCREV` is what remained of another path to remove this package, I'll undo it.
And the version used in our master is 2.12, so there's no need to adapt the patches for 2.11.

EdTheBearded: I agree, pinning a hash would not be nice, and would only lead to us having something outdated.
Usually, when we ping a hash, we do so with AUTOREV on `use-head-next`, so our integration build keep pulling the latest from upstream, and we manually bump before every release. Also done in recipes we maintain, never in a .bbappend.

rborn-tx: > And the version used in our master is 2.12, so there's no need to adapt the patches for 2.11.

@Freireg Am I missing something? Are we not in 2.11 in the manifest?

MatheusRodrigues-tor (approved):

recipes-support/libbytesize/libbytesize_%.bbappend
Freireg: @rborn-tx the `libbytesize` version is pinned by `meta-oe` which points to the 2.12 release in both Master and Wrynose (https://layers.openembedded.org/layerindex/recipe/68568/).

rborn-tx: @Freireg Here's how you would check it:

Our manifest, in its `master` branch points to `meta-openembedded` @ `b22026cd3785700052aefa90e709b92d8e7f74fc`:

https://gitlab.com/toradex/rd/linux-bsp/toradex-manifest/-/blob/master/base/pinned.xml?ref_type=heads#L9

That version of `meta-openembedded` refers to `libbytesize` version 2.11:

https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/libbytesize/?h=b22026cd3785700052aefa90e709b92d8e7f74fc

We are not using the latest version of `meta-openembedded` (where the layer `meta-oe` is kept); that's why we're not yet on 2.12.

EdTheBearded: Hi @rborn-tx. I think you're mixing the manifests.
On master we only use the `next.xml`, which does not include any of the `pinned` manifests, but the `integration` ones.
That's why the version mismatch.
I just checked here the master I synced some days ago, and indeed it's 2.12:
```
fee@threadripper2:/build/fee/torizon-integration-master$ find . -name "libbytesize*"
./layers/meta-openembedded/meta-oe/recipes-support/libbytesize
./layers/meta-openembedded/meta-oe/recipes-support/libbytesize/libbytesize_2.12.bb
```

EdTheBearded:
> Nitpick: typo in commit message: `dependecy`

There's still another type in the message body :grinning:

EdTheBearded (approved):

jsrc27 (approved):

recipes-support/libbytesize/libbytesize_%.bbappend
rborn-tx: Oh Gosh, sorry for that guys. I was checking the integration manifest for the master branch all the time.

I knew that `next.xml` was used only by the Ext Integration matrices on Jenkins so I didn't even bother to check the actual matrices. But now I did and saw that `master` only has an Ext Integration matrix (and no nightly matrix).

rborn-tx (approved):

PR #551, opened 2026-06-18
apparmor: Add apparmor userspace utils (master)

jsrc27:
Same as: https://github.com/torizon/meta-toradex-torizon/pull/548

But, for the `master` branch. This doesn't have the patch that was needed on Scarthgap since that patch was originally backported from `master` anyways.

### DIFF
--- a/conf/distro/include/base-distro.inc
+++ b/conf/distro/include/base-distro.inc
@@ -38,7 +38,7 @@ IMAGE_BOOT_FILES = ""
 IMAGE_BOOT_FILES_REMOVE:sota = "boot.scr-${MACHINE};boot.scr overlays.txt overlays/*;overlays/"
 IMAGE_BOOT_FILES_REMOVE:append:apalis-imx8 = " hdmitxfw.bin dpfw.bin"
 
-DISTRO_FEATURES:append = " virtualization stateless-system"
+DISTRO_FEATURES:append = " virtualization stateless-system apparmor"
 DISTRO_FEATURES_REMOVE ?= "3g alsa irda pcmcia nfc ldconfig pulseaudio wayland x11 ptest multiarch vulkan"
 DISTRO_FEATURES_REMOVE:remove:verdin-imx95 = "vulkan"
 DISTRO_FEATURES_REMOVE:remove:toradex-smarc-imx95 = "vulkan"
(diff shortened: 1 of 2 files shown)


jsrc27:
One note for the team here. Adding the `apparmor` package does pull in a `bash` dependency as seen here:

```
$ bitbake -e apparmor | grep ^RDEPENDS:apparmor=
RDEPENDS:apparmor=" glibc-utils python3-core bash  python3-core python3-modules"
```

However, as far as I can tell this is because the recipe inherits `bash-completion`: https://git.yoctoproject.org/meta-security/tree/recipes-mac/AppArmor/apparmor_4.0.3.bb?h=master#n31

In the future, once we disable/remove `bash-completion` in our layer the dependency here should go away as well.

rborn-tx:
> One note for the team here. Adding the `apparmor` package does pull in a `bash` dependency as seen here:
> 
> ```
> $ bitbake -e apparmor | grep ^RDEPENDS:apparmor=
> RDEPENDS:apparmor=" glibc-utils python3-core bash  python3-core python3-modules"
> ```
> 
> However, as far as I can tell this is because the recipe inherits `bash-completion`: https://git.yoctoproject.org/meta-security/tree/recipes-mac/AppArmor/apparmor_4.0.3.bb?h=master#n31

Checking `apparmor_4.0.3.bb` I see:

```
PACKAGECONFIG ?= "python perl aa-decode"
...
PACKAGECONFIG[aa-decode] = ",,,bash"
```

And checking the `apparmor` source repository, I see that `aa-decode` is indeed a Bash script. So, I'd imagine that Bash is being pulled in by the presence of `aa-decode` in its `PACKAGECONFIG`, unless we are changing that default somewhere (I didn't check).

jsrc27:
> So, I'd imagine that Bash is being pulled in by the presence of aa-decode in its PACKAGECONFIG

Oh true, I missed that config, so that would be pulling in bash as well. I suppose let's leave it in for now, and I'll add a note to the existing ticket for removing `bash`. That way, all the bash-removal work and discussion stays organized in one place. If that's okay with everyone.

rborn-tx (approved):

EdTheBearded (approved):

EdTheBearded:
@jsrc27 Could you please refactor this commit?
This morning our common builds were failing due to this .bbappend. Since we don't pull the meta-security layer for them, this .bbappend is hanging.

~I'll submit a fix for that, putting this guy in a dynamic-layer.~
PR: https://github.com/torizon/meta-toradex-torizon/pull/552

MatheusRodrigues-tor (approved):

Freireg (approved):

PR #555, opened 2026-06-22
Install AppArmor only on Toradex Devices (master)

jsrc27:
Similar to: https://github.com/torizon/meta-toradex-torizon/pull/553

But for the `master` branch. No need for dynamic layers here so we have no bbappend on `master` for AppArmor.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -25,12 +25,12 @@ BBFILES_DYNAMIC += "\
     meta-synaptics:${LAYERDIR}/dynamic-layers/meta-synaptics/*/*/*.bbappend \
     meta-ti-bsp:${LAYERDIR}/dynamic-layers/meta-ti-bsp/*/*/*.bb \
     meta-ti-bsp:${LAYERDIR}/dynamic-layers/meta-ti-bsp/*/*/*.bbappend \
-    toradex-bsp-common-layer:${LAYERDIR}/dynamic-layers/meta-toradex-bsp-common/*/*/*.bb \
-    toradex-bsp-common-layer:${LAYERDIR}/dynamic-layers/meta-toradex-bsp-common/*/*/*.bbappend \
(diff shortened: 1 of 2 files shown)


MatheusRodrigues-tor (approved):

Freireg (approved):

rborn-tx (approved):

EdTheBearded:
Hi @jsrc27, why you say we don't need dynamic layers here?
My understanding is that we don't see an issue now because we don't build master for Common Torizon devices, but once we go to our next major version we'll see that same error again, no? Making the dynamic layers fix necessary.
Or am I missing something?

jsrc27:
@EdTheBearded 

Wasn't the issue on the Scarthgap branch that we were bbappending onto `apparmor` when the base recipe wasn't present in Common Torizon?

In Master we don't have a bbappend for `apparmor`. Furthermore, with this we're now only installing `apparmor` in cases where the `tdx` override applies. Which would completely exclude Common Torizon images right? So `apparmor` shouldn't even be in the build for Common Torizon builds.

Or is there another reason we would need dynamic layers here?

EdTheBearded:
> @EdTheBearded
> 
> Wasn't the issue on the Scarthgap branch that we were bbappending onto `apparmor` when the base recipe wasn't present in Common Torizon?
> 
> In Master we don't have a bbappend for `apparmor`. Furthermore, with this we're now only installing `apparmor` in cases where the `tdx` override applies. Which would completely exclude Common Torizon images right? So `apparmor` shouldn't even be in the build for Common Torizon builds.
> 
> Or is there another reason we would need dynamic layers here?

Oh I overlooked the .bbappend! That's right, you're already installing this with `tdx` override, and if we don't have the bbappend we're good!

Thanks!!

EdTheBearded (approved):

PR #558, opened 2026-06-25
ostree_layer_revision_info: exclude OSTREE_LAYER_REVISION_INFO from vardeps to prevent conflicting basehashes

Freireg:
This change exclude `OSTREE_LAYER_REVISION_INFO` from `ostreecommit` basehash calculation (which is the part relevant to bitbake's signature/determinism check). The actual value computed at runtime is left unchanged, so the real layer-revision metadata is still committed into the OSTree image as before.

Related-to: TOR-4216

### DIFF
--- a/classes/torizon_base_image_type.inc
+++ b/classes/torizon_base_image_type.inc
@@ -85,6 +85,7 @@ validate_ostree_layer_revision_info() {
         bbfatal "OSTREE_LAYER_REVISION_INFO is not set, but is required for OSTree image commits"
     fi
 }
+validate_ostree_layer_revision_info[vardepsexclude] += "OSTREE_LAYER_REVISION_INFO"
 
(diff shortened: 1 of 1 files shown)


jsrc27 (dismissed):

MatheusRodrigues-tor (dismissed):

classes/torizon_base_image_type.inc
rborn-tx, line 1: @Freireg Regarding the commit messages: could you please follow our commit guidelines (linked below)?

https://github.com/torizon/meta-toradex-torizon/blob/scarthgap/docs/#L28

rborn-tx (changes requested):

classes/torizon_base_image_type.inc
Freireg: @rborn-tx done!

rborn-tx (approved):

MatheusRodrigues-tor (approved):

jsrc27 (approved):

EdTheBearded (approved):

PR #559, opened 2026-06-25
setup-devices: add `use-head-next` to common machines

Freireg:
Adding a logic block to identify if the `common-torizon` build is using the integration manifest and set the `:use-head-next` on the build. This flag is needed to allow some recipes to overwrite the pinned hashes during this type of build.

Related-to: TOR-4450

### DIFF
--- a/scripts/lib/setup-devices/setup-environment-imx
+++ b/scripts/lib/setup-devices/setup-environment-imx
@@ -39,10 +39,37 @@ INHERIT += "rm_work"
 # Torizon
 include conf/machine/include/\${MACHINE}.inc
 EOF
+
+fi
+
+_TDX_MANIFEST_FILE="${_TDX_OEROOT}"/.repo/manifest.xml
+_TDX_INTEGRATION_BUILD=0
+if [ -L "$_TDX_MANIFEST_FILE" ]; then
+  case "$(basename "$(readlink -f "$_TDX_MANIFEST_FILE")")" in
+    integration.xml|next.xml) _TDX_INTEGRATION_BUILD=1 ;;
+  esac
+elif [ -f "$_TDX_MANIFEST_FILE" ]; then
+  if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+    _TDX_INTEGRATION_BUILD=1
+  fi
+fi
+
+# When building on integration, ensure use-head-next is set so you always get
+# the newest software (u-boot, kernel, aktualizr-torizon, etc). Idempotent, so
+# it also applies when local.conf already existed from a previous run.
+if [ "$_TDX_INTEGRATION_BUILD" = "1" ] && \
+   ! grep -qF 'DISTROOVERRIDES .= ":use-head-next"' conf/local.conf; then
+  cat <<EOF >>conf/local.conf
+
+# This is needed when building on integration. With use-head-next you
+# always get the newest software (u-boot, kernel, aktualizr-torizon, etc).
+# Building on integration without use-head-next your build may fail.
+DISTROOVERRIDES .= ":use-head-next"
+EOF
 fi
 
 
-if ! grep -q "# Torizon" conf/bblayers.conf; then 
+if ! grep -q "# Torizon" conf/bblayers.conf; then
   cat <<EOF >>conf/bblayers.conf
 # Torizon
 BBLAYERS += "\${BSPDIR}/sources/meta-toradex-torizon"
--- a/scripts/lib/setup-devices/setup-environment-intel
+++ b/scripts/lib/setup-devices/setup-environment-intel
@@ -31,7 +31,35 @@ INHERIT += "rm_work"
 # Torizon
 include conf/machine/include/\${MACHINE}.inc
 EOF
+
 fi
+
+_TDX_MANIFEST_FILE="${_TDX_OEROOT}"/.repo/manifest.xml
+_TDX_INTEGRATION_BUILD=0
+if [ -L "$_TDX_MANIFEST_FILE" ]; then
+  case "$(basename "$(readlink -f "$_TDX_MANIFEST_FILE")")" in
+    integration.xml|next.xml) _TDX_INTEGRATION_BUILD=1 ;;
+  esac
+elif [ -f "$_TDX_MANIFEST_FILE" ]; then
+  if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+    _TDX_INTEGRATION_BUILD=1
+  fi
+fi
+
+# When building on integration, ensure use-head-next is set so you always get
+# the newest software (u-boot, kernel, aktualizr-torizon, etc). Idempotent, so
+# it also applies when local.conf already existed from a previous run.
+if [ "$_TDX_INTEGRATION_BUILD" = "1" ] && \
+   ! grep -qF 'DISTROOVERRIDES .= ":use-head-next"' conf/local.conf; then
+  cat <<EOF >>conf/local.conf
+
+# This is needed when building on integration. With use-head-next you
+# always get the newest software (u-boot, kernel, aktualizr-torizon, etc).
+# Building on integration without use-head-next your build may fail.
+DISTROOVERRIDES .= ":use-head-next"
+EOF
+fi
+
 
 if ! grep -q "# Torizon" conf/bblayers.conf; then
   cat <<EOF >>conf/bblayers.conf
--- a/scripts/lib/setup-devices/setup-environment-nvidia
+++ b/scripts/lib/setup-devices/setup-environment-nvidia
@@ -146,6 +146,22 @@ BBLAYERS = " \\
 EOF
 fi
 
+_TDX_MANIFEST_FILE="${_TDX_OEROOT}"/.repo/manifest.xml
+if [ ! -e "$_TDX_MANIFEST_FILE" ]; then
+    _TDX_MANIFEST_FILE="${_TDX_OEROOT}"/.repo/manifests/manifest.xml
+fi
+_TDX_INTEGRATION_BUILD=0
+
+if [ -L "$_TDX_MANIFEST_FILE" ]; then
+    case "$(basename "$(readlink -f "$_TDX_MANIFEST_FILE")")" in
+        integration.xml|next.xml) _TDX_INTEGRATION_BUILD=1 ;;
+    esac
+elif [ -f "$_TDX_MANIFEST_FILE" ]; then
+    if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+        _TDX_INTEGRATION_BUILD=1
+    fi
+fi
+
 if [ "$_TDX_AUTO_CONF_CREATED" = "0" ]; then
     cat >> conf/auto.conf <<EOF
 DISTRO ?= "${DISTRO}"
@@ -159,6 +175,21 @@ else
     sed -i "s/^DISTRO ?= \"[^\"]*\"$/DISTRO ?= \"${DISTRO}\"/" conf/auto.conf
     sed -i "s/^MACHINE ?= \"[^\"]*\"$/MACHINE ?= \"${MACHINE}\"/" conf/auto.conf
     sed -i "s/^SDKMACHINE ?= \"[^\"]*\"$/SDKMACHINE ?= \"${SDKMACHINE}\"/" conf/auto.conf
+fi
+
+# When building on integration (integration.xml/next.xml), ensure use-head-next
+# is set so you always get the newest software (u-boot, kernel,
+# aktualizr-torizon, etc). Prepended to the top of auto.conf, and applied even
+# when auto.conf already existed from a previous run.
+if [ "$_TDX_INTEGRATION_BUILD" = "1" ] && \
+   ! grep -qF 'DISTROOVERRIDES .= ":use-head-next"' conf/auto.conf; then
+    cat <<EOF >>conf/auto.conf
+
+# This is needed when building on integration. With use-head-next you
+# always get the newest software (u-boot, kernel, aktualizr-torizon, etc).
+# Building on integration without use-head-next your build may fail.
+DISTROOVERRIDES .= ":use-head-next"
+EOF
 fi
 
 cat > conf/site.conf <<_EOF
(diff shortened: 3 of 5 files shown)


scripts/lib/setup-devices/setup-environment-nvidia
```diff
@@ -146,6 +146,22 @@ BBLAYERS = " \\
...
+_TDX_MANIFEST_FILE="${_TDX_OEROOT}"/.repo/manifest.xml
+if [ ! -e "$_TDX_MANIFEST_FILE" ]; then
+    _TDX_MANIFEST_FILE="${_TDX_OEROOT}"/.repo/manifests/manifest.xml
+fi
+_TDX_INTEGRATION_BUILD=0
+
+if [ -L "$_TDX_MANIFEST_FILE" ]; then
+    case "$(basename "$(readlink -f "$_TDX_MANIFEST_FILE")")" in
+        integration.xml|next.xml) _TDX_INTEGRATION_BUILD=1 ;;
+    esac
```
EdTheBearded, line 158: Out of curiosity, why you did this logic differently for nvidia and ti?

Freireg: I started with the same logic block across all scripts, then adjusted as needed. The `nvidia` and `ti` scripts were the last ones I worked on, the generic block wasn't working properly for them so I slightly changed the approach (which seems to be a bit cleaner). Do you think I should rework the other scripts to keep all of them consistent?

scripts/lib/setup-devices/setup-environment-imx
```diff
@@ -39,10 +39,33 @@ INHERIT += "rm_work"
...
+       [ "$(basename `readlink -f "$_TDX_MANIFEST_FILE"`)" == "next.xml" ]; then
+      _TDX_INTEGRATION_BUILD=1
+    fi
+  else
+    if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+      _TDX_INTEGRATION_BUILD=1
+    fi
+  fi
+
+  if [ "$_TDX_INTEGRATION_BUILD" = "1" ]; then
```
rborn-tx: Doesn't this need a protection to prevent the addition of the `DISTROVERRIDES` block multiple times to `local.conf` (as you did for `setup-environment-nvidia`) or is there anything preventing this code from being reached when `local.conf` already exists?

scripts/lib/setup-devices/setup-environment-intel
```diff
@@ -31,6 +31,29 @@ INHERIT += "rm_work"
...
+       [ "$(basename `readlink -f "$_TDX_MANIFEST_FILE"`)" == "next.xml" ]; then
+      _TDX_INTEGRATION_BUILD=1
+    fi
+  else
+    if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+      _TDX_INTEGRATION_BUILD=1
+    fi
+  fi
+
+  if [ "$_TDX_INTEGRATION_BUILD" = "1" ]; then
```
rborn-tx: Same question here: doesn't this need a protection to prevent the addition of the `DISTROVERRIDES` block multiple times to `local.conf`?

scripts/lib/setup-devices/setup-environment-nvidia
```diff
@@ -161,6 +177,23 @@ else
...
+    { cat <<EOF
+# This is needed when building on integration. With use-head-next you
+# always get the newest software (u-boot, kernel, aktualizr-torizon, etc).
+# Building on integration without use-head-next your build may fail.
+DISTROOVERRIDES .= ":use-head-next"
+
+EOF
+      cat conf/auto.conf
+    } > conf/auto.conf.tmp && mv conf/auto.conf.tmp conf/auto.conf
+fi
```
rborn-tx, line 193: Could you clarify why here we prepend whereas with all other machines we append this block?

scripts/lib/setup-devices/setup-environment-synaptics
```diff
@@ -127,6 +127,29 @@ INHERIT += "rm_work"
...
+       [ "$(basename `readlink -f "$_TDX_MANIFEST_FILE"`)" == "next.xml" ]; then
+      _TDX_INTEGRATION_BUILD=1
+    fi
+  else
+    if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+      _TDX_INTEGRATION_BUILD=1
+    fi
+  fi
+
+  if [ "$_TDX_INTEGRATION_BUILD" = "1" ]; then
```
rborn-tx: Ditto

scripts/lib/setup-devices/setup-environment-ti
```diff
@@ -120,6 +120,36 @@ INHERIT += "rm_work"
...
+elif [ -f "$_TDX_MANIFEST_FILE" ]; then
+  if grep -qE "(integration.xml|next.xml)" "$_TDX_MANIFEST_FILE"; then
+    _TDX_INTEGRATION_BUILD=1
+  fi
+fi
+
+# When building on integration, ensure use-head-next is set so you always get
+# the newest software (u-boot, kernel, aktualizr-torizon, etc). Idempotent, so
+# it also applies when local.conf already existed from a previous run.
+if [ "$_TDX_INTEGRATION_BUILD" = "1" ] && \
```
rborn-tx, line 141: Here's the protection I'm talking about that is not present in the other places.

rborn-tx (commented):
Asked some questions for clarification and potential changes.

scripts/lib/setup-devices/setup-environment-imx
Freireg: For the imx, intel and synaptics scripts there's no need for the extra guard since this block is added inside a `if ! grep -q "# Torizon" conf/local.conf` validation, so it only runs once during the initial setup.
The nvidia script needs the explicit check because there the integration block writes to `conf/auto.conf` and is not wrapped in that one-time guard.

The ti's script checks it intentionally outside this if block, since it inherited the same approach I used for nvidia. That leaves two possible paths for consistency:

1. Keep the one-time validation and align ti with imx/intel/synaptics (move the check inside the guard).
2. Update the others to match ti/nvidia, always checking whether `:use-head-next` is already present and only adding it if missing (if it is an integration build).

scripts/lib/setup-devices/setup-environment-nvidia
Freireg: Just to keep consistency with the `auto.conf` generated in our torizon template, where this key is added at the begining of the file.

rborn-tx: Oh, I've just realized that all the other setup-environment scripts being introduced are adding `use-head-next` to `auto.conf` except this one (nvidia). Is there a reason for that?

scripts/lib/setup-devices/setup-environment-imx
rborn-tx: The second approach seems more robust to me.

scripts/lib/setup-devices/setup-environment-nvidia
rborn-tx: Regarding the consistency:

Since the location of the additional line inside `auto.conf` doesn't matter, I'd go with appending since it's slightly simpler. Also, in the torizon template, we create the file from scratch which happens to place the assignment at the beginning of the file, that is, we don't implement special logic for prepending there either.

Freireg: > Oh, I've just realized that all the other setup-environment scripts being introduced are adding use-head-next to auto.conf except this one (nvidia). 

Don't you mean the opposite? Only nvidia's script uses the `auto.conf`, the others are using `local.conf`. 

> Is there a reason for that?

I'm not sure why but nvidia's setup is the only one using the `auto.conf` since its introduction here: 456fc24a4ed405541dd2c1be3b0653be99458ad8

scripts/lib/setup-devices/setup-environment-nvidia
EdTheBearded: No, in the end it's fine. Just got curious if there was a reason behind, and there is.
All good

scripts/lib/setup-devices/setup-environment-nvidia
rborn-tx: > Don't you mean the opposite? Only nvidia's script uses the auto.conf, the others are using local.conf.

Oh, sorry. Exactly.

> I'm not sure why but nvidia's setup is the only one using the auto.conf since its introduction

Interesting. This was done by a third party and I guess we didn't pay attention to it.

---

Regardless of the "statistics" :grin:, maybe what we should think about is where the `use-head-next` override fits best and use that unless constraints prevent it. For Toradex devices, it looks like we are using `auto.conf` for that purpose:

https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/scripts/lib/setup-devices/setup-environment-toradex#L130

From https://docs.yoctoproject.org/scarthgap/overview-manual/concepts.html#user-configuration:

> auto.conf: The file is usually created and written to by an autobuilder. The settings put into the file are typically the same as you would find in the conf/local.conf or the conf/site.conf files.

So, it seems `auto.conf` is to be used by automated/CI systems (machine-generated stuff). I'd say that `local.conf` is mainly for user-editable variables.

Given that we want `use-head-next` depending on what **manifest** was selected by the user, I'd imagine we don't want users to see/edit the corresponding setting normally. So, the `auto.conf` file seems the best place to add the override, like the Toradex setup script does. Therefore, unless there's some complication with it for a particular vendor I believe `auto.conf` should be the target for the `use-head-next` configuration.

@EdTheBearded @jsrc27 What do you guys think?

MatheusRodrigues-tor (approved):

scripts/lib/setup-devices/setup-environment-nvidia
jsrc27: I think with that rationale `auto.conf` is reasonable. Especially since it sounds like it's not intended for users to really touch/change this `use-head-next` configuration, correct?

scripts/lib/setup-devices/setup-environment-imx
rborn-tx: Looks like this is done. Closing this conversation.

scripts/lib/setup-devices/setup-environment-synaptics
rborn-tx: Looks like this is done.

scripts/lib/setup-devices/setup-environment-nvidia
EdTheBearded: I also agree, but this is something that I could also see as a technical debt.
My take here is that we need a whole refactor of the setup scripts. I did something quick in the past for 2 devices only, and now we have a bunch more. It could be a lot better!
This is why I agree, but when he asked me I said that I'm fine if we actually fix things later with a refactor.
But if you guys do see value in actually organizing things a bit now, I'll agree with you.

rborn-tx: I asked AI to make a table for us to have a clearer picture of where each variable is being set by each setup script (before this PR):


| Variable                 | `toradex`    | `nvidia`     | `stm32mp`    | `ti`         | `intel`      | `renesas`    | `imx`        | `synaptics`  |
|--------------------------|--------------|--------------|--------------|--------------|--------------|--------------|--------------|--------------|
| `DISTRO`                 | `auto.conf`  | `auto.conf`  | `auto.conf`  | `local.conf` | `local.conf` | `local.conf` | `local.conf` | `local.conf` |
| `MACHINE`                | `auto.conf`  | `auto.conf`  | `auto.conf`  | `local.conf` | `local.conf` | `local.conf` |              | `local.conf` |
| `DISTROOVERRIDES`        | `auto.conf`  |              |              |              |              |              |              |              |
| `INHERIT`                | `auto.conf`  | `auto.conf`  | `auto.conf`  | `local.conf` | `local.conf` | `local.conf` | `local.conf` | `local.conf` |
| `SDKMACHINE`             | `auto.conf`  | `auto.conf`  | `auto.conf`  |              |              |              |              |              |
|                          |              |              |              |              |              |              |              |              |
| `ACCEPT_FSL_EULA`        | `local.conf` |              |              |              |              |              |              |              |
| `ACCEPT_EULA_${MACHINE}` |              | `local.conf` | `local.conf` |              |              |              |              |              |
|                          |              |              |              |              |              |              |              |              |
| `DL_DIR`                 | `site.conf`  | `site.conf`  | `site.conf`  | `local.conf` | `local.conf` | `local.conf` | `local.conf` | `local.conf` |
| `DEPLOY_DIR`             | `site.conf`  | `site.conf`  | `site.conf`  |              | `local.conf` | `local.conf` | `local.conf` | `local.conf` |
| `SCONF_VERSION`          | `site.conf`  | `site.conf`  | `site.conf`  |              |              |              |              |              |
| `SSTATE_DIR`             | `site.conf`  | `site.conf`  | `site.conf`  | `local.conf` | `local.conf` | `local.conf` | `local.conf` | `local.conf` |
| `TI_COMMON_DEPLOY`       | `site.conf`  |              |              | `local.conf` | `local.conf` |              | `local.conf` |              |
| `_COMMON_DEPLOY`         |              | `site.conf`  | `site.conf`  |              |              |              |              |              |
| `TMPDIR`                 | `site.conf`  | `site.conf`  | `site.conf`  | `local.conf` | `local.conf` | `local.conf` | `local.conf` | `local.conf` |
|                          |              |              |              |              |              |              |              |              |
| `LICENSE_FLAGS_ACCEPTED` |              |              |              |              | `local.conf` |              |              | `local.conf` |
| `OPTEE_TA_ENC`           |              |              |              |              |              |              |              | `local.conf` |
| `SOURCE_MIRROR_URL`      |              |              |              |              |              |              |              | `local.conf` |


We can see that the `nvidia` and `stm32mp` are aligned with the `toradex` script (which I'm considering the the "golden standard" / reference). The others are placing all variables in `local.conf`. So, if we wanted to be consistent across all vendors, it seems we'd need to change many things.

Given that, aligning with the reference only for the `DISTROOVERRIDES` variable doesn't seem to add lots of value. So, I'm okay with leaving this as a technical debt.

@Freireg Please few free to resolve this conversation once you read this.

rborn-tx (approved):

scripts/lib/setup-devices/setup-environment-nvidia
EdTheBearded: Technical debt registered :grin: 
https://toradex.atlassian.net/browse/TOR-4511

Thanks @griloHBG!

EdTheBearded (approved):

jsrc27 (approved):

PR #568, opened 2026-07-07
kernel-devsrc: Backport fix for 6.12+ kernels

jsrc27:
Using the deployed kernel source for compiling kernel modules fails if the Linux kernel version is v6.12 or higher.

There is a fix for this on upstream:
https://git.openembedded.org/openembedded-core/commit/meta/recipes-kernel/linux/kernel-devsrc.bb?id=b3c24a31c29aa74a9d63a0ea0bcaccca73db870b

Backport this for the sl1680 machines where the kernel version is v6.12 at the time of writing.

Related-to: TCB-867

### DIFF
--- a/recipes-kernel/linux/kernel-devsrc.bbappend
+++ b/recipes-kernel/linux/kernel-devsrc.bbappend
@@ -1,4 +1,19 @@
 do_install:append() {
+    (
+        cd ${S}
+
+        if [ "${ARCH}" = "arm64" ]; then
+            # 6.12+
+            cp -a --parents arch/arm64/kernel/vdso/vgetrandom.c $kerneldir/build/ 2>/dev/null || :
+            cp -a --parents arch/arm64/kernel/vdso/vgetrandom-chacha.S $kerneldir/build/ 2>/dev/null || :
+
+            cp -a --parents arch/arm64/tools/syscall_64.tbl $kerneldir/build/ 2>/dev/null || :
+            cp -a --parents arch/arm64/tools/syscall_32.tbl $kerneldir/build/ 2>/dev/null || :
+
+            chown -R root:root ${D}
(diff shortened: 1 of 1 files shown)


recipes-kernel/linux/kernel-devsrc.bbappend
```diff
@@ -1,3 +1,20 @@
+do_install:append:sl1680() {
```
jsrc27: So I wanted to be somewhat conservative with the scope of my change and only apply this to the sl1680 machines.

But looking at the upstream recipe I guess there's no harm in having this in the generic `do_install`. It's a relatively minor detail, but let me know if this is preferable. Or if we're okay as it is now with the sl1680 specific `do_install`.

EdTheBearded: I think it is okay, specially if we don't have any complaints with the other machines.
But shouldn't we also include sl2619?

jsrc27: > But shouldn't we also include sl2619?

Ahh that's true, I assume this also uses the newer 6.12 kernel as well. I'll add another `do_install:append` targeting this machine.

jsrc27: Done in V2.

EdTheBearded (dismissed):

jsrc27:
Turns out several of the common Torizon machines for TI are also on kernel V6.12+ in Scarthgap. Since this affects multiple machines I've decided to just move the fix to the general `do_install:append`.

Freireg (approved):

MatheusRodrigues-tor (approved):

EdTheBearded (approved):

PR #570, opened 2026-07-13
lec-mtk-i1200: add ADLINK LEC-MTK-i1200 (MediaTek Genio 1200) board support

vdskvortsov:
This series adds Common Torizon OS board support for the ADLINK LEC-MTK-i1200 SMARC module (MediaTek Genio 1200 / MT8395), on top of the MediaTek IoT Yocto (Genio) BSP layers.

Commits:

- lec-mtk-i1200: add build environment for the ADLINK Genio 1200
  Registers the meta-mediatek-bsp and meta-adlink-mtk dynamic layers, maps MACHINE=lec-mtk-i1200-ufs in setup-environment, adds the setup-environment-mtk helper, and fixes vendor BSP build failures so the tree builds cleanly.

- lec-mtk-i1200: integrate Torizon OSTree/U-Boot boot flow
  Boots the standard Torizon OSTree way (distro_bootcmd -> /boot.scr -> OSTree-managed uEnv.txt): machine includes, the OSTree WIC layout, the fitImage config-node/fdtfile wiring, legacy boot.scr support, and /etc/fw_env.config for U-Boot env access and OTA rollback state.

- lec-mtk-i1200: fix non-deterministic do_image_ostreecommit basehash
  Excludes provenance-only OSTREE_LAYER_REVISION_INFO from the do_image_ostreecommit task signature so the basehash is stable across parses.

- docs: lec-mtk-i1200: add build guide
  How to build and flash Common Torizon OS for the board.

- lec-mtk-i1200: host-side TorizonCore Builder wrapper for genio-flash
  A host-side wrapper that applies a torizoncore-builder rootfs customisation to a genio-flash image, handling the 4096-byte-sector UFS WIC.

Validation: built for MACHINE=lec-mtk-i1200-ufs and validated on hardware. Boot, UFS storage, USB host, GPIO, both Ethernet interfaces, cloud connectivity, OTA update, and the TorizonCore Builder flow pass. Open items on the reference carrier: one carrier serial port and HDMI output.

## Summary by CodeRabbit

* **New Features**
  * Added Common Torizon OS support for the MediaTek Genio 1200-based ADLINK LEC-MTK-i1200.
  * Added OTA-capable UFS boot configuration, stable USB Wi-Fi naming, TPM2 support, and RTL8821AU Wi-Fi support.
  * Added image conversion tools for customizing, repackaging, and flashing Genio images.
  * Improved boot-device detection and GPT partition resizing.

* **Documentation**
  * Added setup, build, flashing, boot, customization, and Wi-Fi naming guidance.
  * Added the board to the supported machine overview.


dynamic-layers/meta-mediatek-bsp/recipes-support/tcb-genio-bridge/files/tcb-genio-bridge
```diff
@@ -0,0 +1,176 @@
...
+# that patches it for the Genio's 4096-byte-sector WIC (see below).
+run_tcb() {
+    local workdir=$1; shift
+    # Stock tcb's raw-image path assumes 512-byte sectors end to end; the Genio
+    # UFS WIC is 4096-byte-sector. This in-container launcher patches tcb before
+    # handing off: (1) open every libguestfs drive with blocksize=4096 so tcb can
+    # read the 4K GPT; (2) build the output in place — copy the base (partition
+    # layout preserved for genio-flash) and let tcb reformat the existing rootfs
+    # partition — instead of tcb's virt-resize step, an external 512-only tool
+    # that cannot read the 4K disk.
```
jsrc27: In terms of maintenance, have you considered opening a PR On the TorizonCore Builder side to add an option to adjust the expected block-size for raw images?

The patching here is a bit fragile since TorizonCore Builder could of course change in the future, which could break this.

conf/machine/include/common-mtk.inc
```diff
@@ -0,0 +1,82 @@
...
+# OSTREE_LAYER_REVISION_INFO records each layer's branch:sha as OSTree commit
+# metadata (provenance), but it also feeds do_image_ostreecommit's task hash via
+# the validate prefunc. Because it reflects live layer git state, a layer whose
+# checkout is not stable across parses makes the basehash change every parse: the
+# task never hits sstate and the image rebuilds needlessly. This surfaced here
+# with the manually-cloned BSP layer checkouts (meta-mediatek-bsp /
+# meta-adlink-mtk). Exclude it from the hash; it is still computed and written
+# as metadata.
+validate_ostree_layer_revision_info[vardepsexclude] += "OSTREE_LAYER_REVISION_INFO"
+do_image_ostreecommit[vardepsexclude] += "OSTREE_LAYER_REVISION_INFO"
```
jsrc27: There's another PR open to address this issue: https://github.com/torizon/meta-toradex-torizon/pull/558

You could probably drop this from your PR then.

jsrc27 (commented):
Some minor comments

vdskvortsov:
Thanks for the review, @jsrc27 — both points make sense.

**OSTree determinism (`common-mtk.inc`):** Agreed, #558 handles this at the class level and supersedes our per-include lines. I've dropped that commit from this PR.

**Raw-image block size (`tcb-genio-bridge`):** Also agreed — patching TorizonCore Builder from inside the container is fragile. The clean fix is to make TorizonCore Builder itself aware of the raw image's sector size, so the bridge has nothing to patch. I've prototyped that as a `--raw-sector-size` option (with `sector-size:` / `base-sector-size:` tcbuild keys) and validated it end-to-end on the Genio's 4Kn UFS.

On sequencing, here's what I'd propose — happy to follow your preference:
1. Open a PR against `torizon/torizoncore-builder` adding the sector-size option; then, once it's merged and in a released image,
2. Open a follow-up PR here to update `tcb-genio-bridge` to drop the runtime patching and just pass the sector size through.

That leaves the current self-contained bridge in this PR for now (it works against the released `:3` image). Would you prefer that, or that I pull the bridge out of this PR until the TCB option ships?

Separately, an important finding from this work: the Genio UFS is 4Kn (4096-byte logical sectors), and 4K media surfaced two rough issues. One is in TorizonCore Builder's raw-image path, addressed by the sector-size option above. The other is in `resize-helper`, which never grows the rootfs on a 4Kn target: on first boot the partition is enlarged on disk, but while root is mounted the kernel doesn't adopt the new partition size live, so `resize2fs` sees the old size and finds nothing to do — and because the one-shot disables itself afterwards, it never retries, leaving the rootfs stuck at its flashed size. Since `resize-helper` lives in this repo, we'd like to fix it — would you prefer that fix folded into this PR, or as a separate one?

jsrc27:
> That leaves the current self-contained bridge in this PR for now (it works against the released :3 image). Would you prefer that, or that I pull the bridge out of this PR until the TCB option ships?

This sounds reasonable to me. As long as there is a plan to eventually not patch TCB from inside the container. Also the ability to optionally change the block size is probably just a good addition to have in TCB in general.
 
> would you prefer that fix folded into this PR, or as a separate one?

I think it's fine to have it in this PR, as it is related to the port of this new machine. Maybe the other maintainers have differing opinions.

vdskvortsov:
Two updates on this:

**TCB sector-size option (step 1 of the sequencing):** opened torizon/torizoncore-builder#156, which adds the `--raw-sector-size` option (and `sector-size:` / `base-sector-size:` tcbuild keys). Once it merges and ships in a released image, I'll open the follow-up here to drop the runtime patching from `tcb-genio-bridge` and just pass the sector size through.

**resize-helper 4Kn fix:** added as a separate commit here, per your note that it's fine in this PR. It drives the GPT grow through `sgdisk` (which reads the device's logical sector size, so the geometry is correct on 4Kn) instead of the 512-centric scripted `fdisk` edit, and bails out before the destructive recreate if the original start sector can't be read. The MBR path and 512-byte-logical boards are unchanged. Verified on the Genio 4Kn UFS: the rootfs now grows to fill the device.

vdskvortsov:
Pushed two more changes onto this branch:

- **`adapt vendor U-Boot config to the Torizon boot flow`** (new commit, right after the boot-flow commit): raises `SYS_MAXARGS` to 64 so Torizon's `uEnv.txt` `set_bootargs` runs (the vendor default of 16 left `bootcmd_args` undefined), drops the vendor splash (it loads a logo from a bootassets partition Torizon doesn't carry), and disables the FDT env import (the control DTB has no `/config/environment` node). None of these affected boot functionality — they only produced boot-log errors, which the changes clean up.
- **GPT attribute preservation** (folded into the resize-helper commit): the `sgdisk` delete/recreate was zeroing the partition's GPT attributes, clearing the legacy-BIOS-bootable flag (`--active`) that U-Boot's `distro_bootcmd` scan uses to find the rootfs — so the board dropped to the U-Boot prompt after the first post-resize reboot. The attributes are now captured and restored across the resize.

rborn-tx (dismissed):
@vdskvortsov I went through the various commits and overall they look great.

The only thing that doesn't look right is the commit:

```
lec-mtk-i1200: host-side TorizonCore Builder wrapper for genio-flash
```

I believe the goal of this commit was to allow users to customize `MediaTek Genio` images via TorizonCore Builder, right?

However, in general we assume that users of TorizonCore Builder are not Yocto users. Because of that, invoking that tool from Yocto wouldn't be the proper/expected approach.

I'll soon make a separate comment to explain what we did for the Synaptics Astra machine. That machine falls into a very similar case as we have here where the build output does not produce a standard WIC or Toradex Easy Installer image which are the only formats supported by TorizonCore Builder at the moment.

rborn-tx:
### Integration of the Synaptics Astra machine with TorizonCore Builder (for reference)

With the Synaptics Astra machine, what we did was we created conversion scripts from/to the Synaptics image format and the WIC format. The scripts are implemented and have some initial documentation in this repository:

https://gitlab.com/toradex/rd/torizon-core/syna2img

This is referenced in the following recipe in layer `meta-toradex-torizon`:

https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/recipes-support/syna2img-tools/syna2img-tools_git.bb

When an image for Synaptics Astra is built, the conversion scripts are shipped along with the image in a tarball, as you can see here:

https://github.com/torizon/meta-toradex-torizon/blob/scarthgap-7.x.y/conf/machine/include/sl-common.inc#L82

If users want to customize the image with TorizonCore Builder, they would convert the image to WIC using one of the scripts, then make any customization they want and, at the end use the other script to convert back to the original format so the image can be used with standard Synaptics tooling.

Notice the procedure performed by the user does not involve Yocto at all.

By the way, this workflow is somewhat new and not yet documented. At the time of this writing, the only documentation we have is the README in the repository I mentioned at the beginning.

---

In the present PR, I'd say we'd like to follow the same approach. That means, we'd have new conversion scripts being deployed alongside the output image. Details would likely need to be discussed with the other maintainers of the present layer...

dmitrys-emcraft:
Replaced the wrapper with simple image converters and added several changes needed for lec-mtk-i1200:

- tcb-genio-bridge: replace the script with genio2img/img2genio converters
  Replaces the wrapper with two host-side converters (aiotflash.tar <-> raw image) and modifies lec-mtk-i1200-ufs.inc to place them beside the Torizon OS image, similar to Synaptics Astra.
  TCB now runs directly against the raw image, but now depends on https://github.com/torizon/torizoncore-builder/pull/156

- aktualizr: carry /var/sota into torizoncore-builder-customised images
  A customised image previously lost the update client's state directory, now recreated via tmpfiles.

- lec-mtk-i1200: deliver Torizon's kernel fragment past the vendor clobber
  The vendor's do_copy_defconfig was discarding every Yocto-merged kernel config fragment.

- lec-mtk-i1200: disable Plymouth splash and route console to fbcon
  Routes console output to fbcon and disables the splash so it doesn't obscure the boot messages and login prompt.

- rtw88: add out-of-tree recipe for RTL8821AU USB Wi-Fi
  Backported driver for the RTL8821AU, which has no in-tree support on this kernel.

- usb-wifi-naming: name USB Wi-Fi interfaces from their MAC
  net_id can't distinguish this SoC's four USB controllers, so two adapters' names could collidenames generated from MAC address instead.

dmitrys-emcraft:
Updated usb-wifi-naming: replaced the MAC-based naming with a USB port-based scheme.

- usb-wifi-naming: name USB Wi-Fi interfaces after the SMARC USB socket
  Assigns a persistent name per SMARC USB signal (USB0-USB5), fixed on the module and identical on every carrier and after every boot. MAC-derived naming stays as the catch-all rule for adapters behind a carrier-side hub.

Also rebased the branch onto the current scarthgap-7.x.y.

recipes-sota/aktualizr-torizon/files/aktualizr-tmpfiles.conf
```diff
@@ -1,2 +1,7 @@
 # Aktualizr does not want these files to be removed by automatic cleanup
 x /tmp/aktualizr-* 0700 root root -
+# Nothing on the target recreates aktualizr's state directory, and an image
+# customised by torizoncore-builder does not inherit it from the base deployment.
+# Must stay lower-case 'd': 'D' would empty it on every boot, taking the device
+# key with it.
+d /var/sota 0700 root root -
```
jsrc27, line 7: Could you elaborate more on this? I'm not sure I fully understand the need for this addition.

The commits says `/var/sota` gets lost due to TorizonCore Builder. But, in a base image is there anything in `/var/sota` to begin with?

vladimirkhusainov-emcraft: Yes — a stock image does have `/var/sota`. It is not part of the OSTree commit, which is why a customised image can end up without it. meta-updater's `image_types_ota.bbclass` (~66–87) creates it directly in the deployment's `var` (`mkdir -p`, `chmod 700`, plus `import/installed_versions`). OSTree deploys an empty `/var` and never copies a commit's `/var` into it, so the deployment `var` is the only place the directory exists.

torizoncore-builder's `deploy` creates a new deployment and copies back only `var/rootdirs` from the old one — `copy_files_from_old_sysroot()`, `tcbuilder/backend/deploy.py:217–234`, called from `deploy_ostree_local()`, which runs on the TEZI path as well as raw/WIC. `var/sota` and `var/local` are dropped. We read the deployment `var` out of a customised image to confirm it: it contains `.ostree-selabeled lib lock log rootdirs run tmp` and no `sota`.

Nothing on the target creates it either — the generated `00ostree-tmpfiles.conf` only covers `rootdirs`, `roothome` and `usrlocal`. So `… | sudo tee /var/sota/auto-provisioning.json` fails on a customised image, and a user who creates the directory by hand gets mode 0755, which means any user on the board can enter the directory holding the device's private key.

This came from a customer's feedback, not from a hypothetical: their customisation is overlay-only, so nothing recreated the directory and provisioning stopped at the first step that writes to it.

The proper fix is in torizoncore-builder, but there are two things it cannot do, which is why we think this line belongs in the layer as well. First, an OSTree update never touches `/var`, so a fix in the tool only helps images built after that release, while `d` also corrects the mode of a directory that already exists — so a device already in the field is repaired on its next boot. Second, it works no matter which torizoncore-builder version produced the image. It is also the same mechanism meta-updater already uses for `/var/rootdirs`, `/var/roothome` and `/var/usrlocal`.

conf/machine/include/common-mtk.inc
vladimirkhusainov-emcraft: Agreed — dropped. The branch is now rebased on a base that includes #558, and the machine-local `vardepsexclude` lines are gone from `common-mtk.inc`; the class-level fix covers it.

dynamic-layers/meta-mediatek-bsp/recipes-support/tcb-genio-bridge/files/tcb-genio-bridge
vladimirkhusainov-emcraft: Yes — that's what we did. torizon/torizoncore-builder#156 adds `--raw-sector-size` (plus the tcbuild `sector-size:` / `base-sector-size:` keys); it merged into `trixie` on 5 August.

The patching is gone from this PR too: `c5555cbe` replaced `tcb-genio-bridge` with `genio2img` / `img2genio`, two host-side converters between `aiotflash.tar` and a plain raw image. torizoncore-builder now runs unmodified on that image.

vladimirkhusainov-emcraft:
@rborn-tx — the swap you asked for is in.

`c5555cbe` replaces the `tcb-genio-bridge` wrapper with `genio2img` / `img2genio`: two standalone host-side converters between the Genio flash tarball and a plain raw image, deployed alongside the built image, following the Synaptics Astra pattern. The wrapper is gone — customising an image no longer involves Yocto at all.

The branch is rebased on the current `scarthgap-7.x.y` head, and `jsrc27`'s threads are all answered. Your review from 22 July predates this — could you re-run it when you get a chance?

recipes-sota/aktualizr-torizon/files/aktualizr-tmpfiles.conf
jsrc27: > so nothing recreated the directory and provisioning stopped at the first step that writes to it.

I see so this is specifically to address the auto-provisioning case correct? Which means this commit is somewhat related to this issue: https://github.com/torizon/meta-toradex-torizon/issues/436

The lack of auto-provisioning support for non-Toradex hardware in this meta-layer is a known limitation. How is the customer in this case doing the provisioning? Do they have some process/script that writes the required provisioning material to `/var/sota`?

Edit: Actually thinking further, we have support for provisioning in TCB for raw/wic based images. So this should not be a problem correct? Or am I misunderstanding?

vladimirkhusainov-emcraft: Agreed, there is already support for provisioning in TCB for raw/wic images.
Three gaps, all of which the suggested fix covers:

1. This customer provisions the other documented way, with no scripts of their
   own: overlay-only customisation, flash, then on the device write
   `/var/sota/auto-provisioning.json` and start `auto-provisioning`. Nothing in
   that flow runs `images provision`, so the directory is simply absent.
2. Mode. `mkdir_p` takes no mode argument, so where the tool does create
   `/var/sota` it lands at `0755` with the default umask, rather than the `0700`
   the image class sets. That directory holds the device key and the imported
   credentials, so every user on the board can now traverse into it.
3. Devices already in the field carrying 2. They are provisioned and do take OTA
   updates, but nothing in TCB reaches them, since `/var` survives the update.
   `d` corrects the mode on the next boot.

The proper fix for 1 and 2 belongs in TCB itself. Our fix still makes sense, we
believe: it covers 1 and 2 whichever version of the tool produced the image, and
3 can only be fixed from within the OS image.

Not related to #436, though — that is the provisioning data never being
generated; this is the directory not surviving customisation.

jsrc27: > This customer provisions the other documented way,

What other documented way are you specifically referring to here? Referring to our official documentation on provisioning here: https://developer.toradex.com/torizon/torizoncore/production-programming-in-torizon/#installing-provisioning-data

The only methods we document for installing the provisioning data is via Yocto, which is not supported for common Torizon devices as per #436. Or via TCB which we just established is supported for common devices.

I just want to clarify, since it seems odd to add this "fix" for a method that is not officially supported/documented.

vladimirkhusainov-emcraft: https://developer.toradex.com/torizon/torizon-platform/torizon-updates/how-to-use-secure-offline-updates-with-torizoncore/
This bit: "If you don't want to provision the device for online platform access (or are unable to), you can also use TorizonCore Builder to get your shared provisioning data, then extract it into /var/sota/import/ on your device (note that you'll need root access on the device):", etc gets the user into this situation.

Another issue is the mode. The raw/wic provisioning leaves that dir at 0755, which ends up in a scenario where the sensitive data (private key, etc) is readable by everyone on devices in the field. Not a concern?

jsrc27: Oh I see, that section was to describe a method to manually provision a device in a standalone fashion. For quick tests, or evaluation purposes and such. That method was not really meant to be used at-scale. Also that section specifies offline updates only. It makes no reference to auto-provisioning. The example in that section showcases manually writing `shared-data.tar.gz` to `/var/sota/import`. It makes no mention of `auto-provisioning.json`, which is what you said the customer is writing to the device. So I still maintain that the method the customer is doing here is not really an "official" method for auto-provisioning devices.

So let me summarize here. The customer in this case manually copies the provisioning data to the device. But they want `/var/sota` to exist by default? Just to save the extra step of creating the directory even though they go through the trouble of manually writing `auto-provisioning.json` and running the `auto-provisioning` script?

The change to the `tmpfiles.conf` is not problematic on it's own I suppose. I just find it strange that the purpose of the change is to assist in an "unofficial" method.

@rborn-tx Perhaps you have some thoughts on this point.

> Another issue is the mode. The raw/wic provisioning leaves that dir at 0755, which ends up in a scenario where the sensitive data (private key, etc) is readable by everyone on devices in the field. Not a concern?

This is a separate point to be discussed.

vladimirkhusainov-emcraft: That customer is porting their application to Torizon OS. Their scenario must have been some manual steps during their development but still a support issue that had to be handled. This issue is generic, not specific to the ADLINK port, so, since the change is harmless on its own, fixing it here saves the same support ticket from other customers later.

The mode issue, if our analysis is correct, is a bigger concern. Indeed, not a good thing to let sensitive data be readable by non-root, on devices in the field.

dmitrys-emcraft:
@EdTheBearded @Freireg @MatheusRodrigues-tor @rborn-tx 

  Following up on this PR. The changes addressing the prior review round (@rborn-tx, 2026-07-22) were pushed on 2026-08-05:

  - tcb-genio-bridge replaced with genio2img/img2genio host-side converters, matching the syna2img pattern described for Synaptics Astra
  - aktualizr: carry /var/sota into torizoncore-builder-customised images
  - lec-mtk-i1200: deliver Torizon's kernel fragment past the vendor clobber
  - lec-mtk-i1200: disable Plymouth splash and route console to fbcon
  - rtw88: add out-of-tree recipe for RTL8821AU USB Wi-Fi
  - usb-wifi-naming: name USB Wi-Fi interfaces after the SMARC USB socket
  - Branch rebased onto the current scarthgap-7.x.y

  No review activity has been recorded since. Could one of you take a look when time permits?

vladimirkhusainov-emcraft:
@rborn-tx — following up specifically on your review of 22 July, which is still the formal blocker on this PR (the outstanding `CHANGES_REQUESTED`).

The change you asked for is in, and it follows the `syna2img` pattern you pointed us at:

`c5555cbe` — *tcb-genio-bridge: replace the script with genio2img/img2genio converters* — drops the Yocto-invoked wrapper and replaces it with two standalone host-side converters between the Genio flash tarball and a plain WIC. They ship alongside the image, exactly as `syna2img-tools` does for Synaptics Astra, so the user's path is: `genio2img` → customise with stock `torizoncore-builder` → `img2genio` → flash with the standard MediaTek tooling. No Yocto anywhere in that flow.

The 4Kn side of it is upstream too: `torizon/torizoncore-builder#156` (`--raw-sector-size`, plus the tcbuild `sector-size:` / `base-sector-size:` keys) merged into `trixie` on 5 August, so plain TCB can open the Genio's 4Kn WIC without any wrapper.

The branch has also been rebased onto current `scarthgap-7.x.y` and carries five further board fixes since your review (commits 8–12).

Could you re-run your review so the `CHANGES_REQUESTED` can be cleared? If the wrapper replacement now looks right to you, that unblocks the other reviewers.

rborn-tx:
@vdskvortsov @jsrc27 Sorry for the delay. I am not receiving notifications from this PR for some reason :weary:. I'll be reviewing/answering you soon.

rborn-tx:
> Could you re-run your review so the CHANGES_REQUESTED can be cleared? If the wrapper replacement now looks right to you, that unblocks the other reviewers.

The wrapper replacement looks good to me. I'd just kindly ask you to squash commits:

* `tcb-genio-bridge: replace the script with genio2img/img2genio converters`
* `lec-mtk-i1200: host-side TorizonCore Builder wrapper for genio-flash`

into one, e.g.:

* `tcb-genio-bridge: add genio2img/img2genio image conversion scripts`

I'll go ahead and clear the `CHANGES_REQUESTED` flag, and review the other changes asap.

rborn-tx (commented):
Clearing previous review flag.

dynamic-layers/meta-adlink-mtk/recipes-bsp/u-boot/files/lec-mtk-i1200-ufs/fw_env.config
```diff
@@ -0,0 +1,3 @@
+# U-Boot env on the UFS boot1 LU (CONFIG_ENV_IS_IN_SCSI, dev 1, offset 0x0, size 0x1000).
+# Device	Offset	Env size
+/dev/sdb	0x0000	0x1000
```
rborn-tx: `/dev/sdb` is not a stable name for the UFS boot LU. Every other board in this layer points `fw_env.config` at an `mmcblk*` node (`dynamic-layers/meta-synaptics/.../sl1680/fw_env.config`, `.../am62/fw_env.config`, `.../imx93-11x11-lpddr4x-frdm/fw_env.config`), which is tied to the controller and therefore stable. `sd*` names are not: they are handed out in SCSI/block probe-completion order across *all* SCSI-class devices, UFS logical units and USB mass storage alike.

If a USB mass-storage device is attached at power-on and its `sd` node is allocated before the UFS boot LU, `fw_setenv` writes 0x1000 bytes at offset 0 of *that* device -- destroying its partition table -- and `fw_printenv` reads garbage, so aktualizr loses the OTA rollback state it manages through libubootenv on every update. In the common case UFS (a platform driver, probed early) wins over USB enumeration, so this is "possible, not typical" rather than guaranteed, and I could not measure the probe race without hardware. But when it does happen the consequence is data loss on the user's device plus a broken rollback path.

Please point the entry at a stable symlink for the boot LU -- a `/dev/disk/by-path/platform-*.ufshci-*-lun-1` path, or a small udev rule creating a named symlink (plus an `RDEPENDS` on it). libubootenv opens whatever path it is given, so this is essentially a one-line change.

dynamic-layers/meta-mediatek-bsp/recipes-kernel/linux/linux-mtk_%.bbappend
```diff
@@ -0,0 +1,107 @@
...
+# on the vendor's own task, which they already ship at two graph positions keyed
+# on MACHINE; an append to a renamed task is silently never called.
+do_configure:prepend:lec-mtk1200() {
+    [ -f ${WORKDIR}/torizon.cfg ] || bbfatal "torizon.cfg absent from ${WORKDIR}:" \
+        "it is missing from SRC_URI, or this is an incremental re-run against an" \
+        "rm_work-reaped WORKDIR (build linux-mtk clean), or UNPACKDIR no longer" \
+        "defaults to WORKDIR"
+    cp ${S}/arch/arm64/configs/${KERNEL_CONFIG_AARCH64} ${B}/.config
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${WORKDIR}/torizon.cfg
+}
```
rborn-tx: This restores only `torizon.cfg`; every other merged fragment is still discarded.

The diagnosis in the commit message is correct -- I read the vendor bbappend (`meta-adlink-mtk/recipes-kernel/linux/linux-mtk_%.bbappend`) and `do_copy_defconfig` does `cp ${S}/arch/arm64/configs/${KERNEL_CONFIG_AARCH64} ${B}/.config` on a task scheduled *after* `do_kernel_configme` for every machine except `absolute-vision`. But this repair reproduces that same clobber and then merges exactly one fragment back, so anything else `do_kernel_configme` had merged is still gone.

Concretely, with composefs enabled: `classes/cfs-signed.bbclass:23` does `DISTROOVERRIDES .= ":cfs-support:cfs-signed"`, and `recipes-kernel/linux/linux-torizon.inc` adds `file://composefs.scc` / `file://composefs-fsverity.scc` under exactly those overrides. On this board those fragments would be merged by configme, wiped by the vendor task, and *not* restored -- while the guard at lines 70-75 only checks `CONFIG_EXT4_FS_SECURITY`, `CONFIG_CFS_BANDWIDTH` and `CONFIG_BLK_DEV_THROTTLING`, all of which come from `torizon.cfg`. The build stays green and the kernel silently lacks composefs support. The same applies to any future Torizon `.scc`, and to any fragment another layer adds to `linux-mtk`.

The vendor already ships the correct ordering for the other machine it supports: `bb.build.addtask('do_copy_defconfig', 'do_kernel_configme', 'do_patch', d)` (the `absolute-vision` branch of their anonymous python). Re-scheduling `do_copy_defconfig` that way for `lec-mtk1200` removes the clobber entirely and makes this re-merge, the guard and the parse-time override check all unnecessary. If reordering the vendor task is considered too invasive, save `${B}/.config` at the end of `do_kernel_configme` and merge *that* onto the defconfig, instead of merging one named fragment.

scripts/lib/setup-devices/setup-environment-mtk
```diff
@@ -0,0 +1,235 @@
...
+# Extra options that can be changed by the user
+INHERIT += "rm_work"
+EOF
+else
+    sed -i "s/^DISTRO ?= \"[^\"]*\"$/DISTRO ?= \"${DISTRO}\"/" conf/auto.conf
+    sed -i "s/^MACHINE ?= \"[^\"]*\"$/MACHINE ?= \"${MACHINE}\"/" conf/auto.conf
+    sed -i "s/^SDKMACHINE ?= \"[^\"]*\"$/SDKMACHINE ?= \"${SDKMACHINE}\"/" conf/auto.conf
+fi
+
+cat > conf/site.conf <<_EOF
```
rborn-tx, line 205: This script is a near-copy of `setup-environment-nvidia` -- which is fine and follows existing precedent in this repo -- but the copy silently drops the integration-manifest block that the original carries at exactly this point, between the `auto.conf` write and `site.conf`:

```
if [ "$_TDX_INTEGRATION_BUILD" = "1" ] && \
   ! grep -qF 'DISTROOVERRIDES .= ":use-head-next"' conf/auto.conf; then
    ... DISTROOVERRIDES .= ":use-head-next" ...
fi
```

Consequence: when this machine is built from an `integration.xml` / `next.xml` manifest -- i.e. in Toradex's own integration builds -- `use-head-next` is never set, so `SRCREV:use-head-next = "${AUTOREV}"` never takes effect for `aktualizr-torizon`, `rac`, `tzn-mqtt`, `linux-toradex-kmeta` and `syna2img-tools`, and the build pins stale revisions. The nvidia script's own comment says such a build "may fail". This will not show up on your bench, only in our CI.

Please restore the `_TDX_MANIFEST_FILE` / `_TDX_INTEGRATION_BUILD` detection and the `DISTROOVERRIDES` append verbatim from `setup-environment-nvidia`.

scripts/lib/setup-devices/setup-environment-mtk
```diff
@@ -0,0 +1,235 @@
...
+EOF
+    else
+        cat >> conf/local.conf <<EOF
+
+# Uncomment to accept the commercial license required by the proprietary Mali
+# DDK (libmali). Not needed for the open-source Panfrost/Mesa GPU provider.
+#LICENSE_FLAGS_ACCEPTED += "commercial"
+EOF
+    fi
+fi
```
rborn-tx: `EULA=1` is a no-op on every invocation after the first. Two independent bugs, both introduced by the fork:

1. `setup-environment-nvidia` keeps `EULA` in the early-return condition (`-a "${EULA}" = "$eulaacpt"`); this copy removed that term (line 92) but added an `EULA`-dependent block *after* the early return. So `MACHINE=... source setup-environment build-x` followed by `EULA=1 MACHINE=... source setup-environment build-x` returns at line 94 and never reaches this block.
2. Even without the early return, the guard here is an unanchored `grep`, and the `EULA`-unset branch writes the *commented* line 208 (`#LICENSE_FLAGS_ACCEPTED += "commercial"`) into `conf/local.conf`. That comment matches the grep, so the accept branch can never run on a build directory that has been configured once.

`conf/template/local.conf` contains no `LICENSE_FLAGS_ACCEPTED` of its own, so the only way to get `commercial` accepted through this script is to set `EULA=1` on the very first invocation for a given build directory. `docs/README-genio.md` never mentions `EULA` at all, so a user who follows the documented steps and then hits the libmali license error has no documented way out other than editing `local.conf` by hand.

Fix: move the EULA block above the checksum early-return (or restore `eulaacpt` to the condition), anchor the guard (`grep -q '^LICENSE_FLAGS_ACCEPTED'`), and mention `EULA=1` in the README's build step.

docs/README-genio.md
```diff
@@ -0,0 +1,315 @@
...
+* `-o OUTPUT_TAR` — repacked tarball (default: `WORKDIR-custom.tar`).
+
+Performing image customization
+------
+Prepare your customization — a `changes/` overlay, a container bundle, or both
+(see Customization classes above).
+
+Deploy the tools and collect them beside the image:
+```
+$$ bitbake tcb-genio-bridge
```
rborn-tx: This `bitbake` invocation is unnecessary, and it re-introduces a Yocto step into a flow the PR description advertises as Yocto-free. `conf/machine/include/lec-mtk-i1200-ufs.inc:22` already has `EXTRA_IMAGEDEPENDS:append = " tcb-genio-bridge"`, and `do_build[recrdeptask] += "do_deploy"` (oe-core `base.bbclass:374`) pulls the recipe's `do_deploy` in as part of the image build -- so `tcb-genio-bridge.tar` is already in the deploy directory when `bitbake torizon-docker` finishes. Please drop the line, or reduce it to a note that the tarball is already there.

docs/README-genio.md
```diff
@@ -0,0 +1,315 @@
...
+
+2. **Customize** with `torizoncore-builder` directly, from inside `work/`:
+```
+$ docker run --rm -v "$PWD":/workdir -v tcb-genio-storage:/storage -v /deploy \
+      -v /var/run/docker.sock:/var/run/docker.sock --net=host \
+      torizon/torizoncore-builder:<release> build --file tcbuild-genio.yaml
+```
+   (Or a native `torizoncore-builder` install, run from the same directory —
+   `input.raw-image.local`/`output.raw-image.local` are relative paths that
+   resolve against wherever it is invoked from.)
```
rborn-tx: There is no such thing as a native `torizoncore-builder` installation -- the tool is distributed only as a container image, and there is no host package or standalone executable to install. Please drop this parenthetical, or reword it to say what was actually meant (e.g. that the container can be run from any working directory, since the `input.raw-image.local` / `output.raw-image.local` paths are relative).

docs/README-genio.md
```diff
@@ -0,0 +1,315 @@
...
+$ cd ..
+$ ./img2genio work
+$ tar xf work-custom.tar
+$ cd torizon-docker-lec-mtk-i1200-ufs-*/
+$ genio-flash system
+```
+
+`work/` can be reused for more than one customization from the same base
+image — rerun steps 2–3 with different overlay/compose content without
+re-running `genio2img`. A preloaded container bundle larger than the base
```
rborn-tx: This does not work as written. `torizoncore-builder`'s `build` aborts when the output file already exists unless `--force` is passed (`tcbuilder/cli/build.py`, "Check if output file already exists and fail if it does"), so the second run stops before doing anything. Please add `--force` (or an `rm output.wic`) to the reuse note.

docs/README-genio.md
```diff
@@ -0,0 +1,315 @@
+MediaTek Genio 1200 (Adlink LEC-MTK-i1200 / I-Pi SMARC 1200)
```
rborn-tx, line 1: This new document is not reachable from the repo front page: the top-level `README.md` has a supported-machines table (lines 23-29) with a row per vendor linking to `docs/README-*.md` (Intel, NVIDIA, NXP, Renesas, ST, Synaptics, TI), and the series adds no row for this one. Please add a MediaTek / ADLINK LEC-MTK-i1200 row pointing at `./docs/README-genio.md`.

dynamic-layers/meta-mediatek-bsp/recipes-support/genio2img-tools/genio2img-tools_1.0.bb
```diff
@@ -0,0 +1,41 @@
+SUMMARY = "Host-side genio/WIC converters for genio-flash images"
```
rborn-tx, line 1: The recipe, its directory and the deployed tarball are all still called `tcb-genio-bridge`, although there is no bridge any more -- the design that name described was removed in `c5555cb`. `genio2img-tools` would mirror `syna2img-tools` and say what the thing actually is.

dynamic-layers/meta-mediatek-bsp/recipes-support/tcb-genio-bridge/tcb-genio-bridge_1.0.bb
```diff
@@ -0,0 +1,41 @@
+SUMMARY = "Host-side genio/WIC converters for genio-flash images"
+DESCRIPTION = "genio2img and img2genio are standalone tools bracketing a \
+released torizoncore-builder run against a MediaTek genio-flash \
+(aiotflash.tar) image: genio2img unpacks and unsparses the tarball's system \
+WIC, and img2genio re-sparses and repacks the result (no Yocto build needed \
+in between). Deployed as a single tcb-genio-bridge.tar archive alongside the \
+image. Ships the tools plus an example tcbuild configuration."
+HOMEPAGE = "https://github.com/EmcraftSystems/meta-toradex-torizon"
```
rborn-tx: `HOMEPAGE` points at the contributor's fork. It should be the upstream repository, or the tools' own repository if you decide to extract them.

dynamic-layers/meta-mediatek-bsp/recipes-support/tcb-genio-bridge/tcb-genio-bridge_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+
+# Host-side tools: deploy as one archive beside the flashing artifacts (not
+# into the rootfs), so it travels with the image as a single file to hand off.
+BRIDGE_DEPLOY = "${WORKDIR}/deploy-stage/${BPN}"
+
+do_deploy() {
+    install -m 0755 ${S}/genio2img ${BRIDGE_DEPLOY}/genio2img
+    install -m 0755 ${S}/img2genio ${BRIDGE_DEPLOY}/img2genio
+    install -m 0644 ${S}/tcbuild-genio.yaml ${BRIDGE_DEPLOY}/tcbuild-genio.yaml
+    tar --numeric-owner -cf ${DEPLOYDIR}/${BPN}.tar -C ${BRIDGE_DEPLOY}/.. ${BPN}
```
rborn-tx: `tar --numeric-owner -cf ...` has no `--sort=name` / `--mtime=...`, so the deployed tarball is not byte-reproducible across builds. Cheap to add.

dynamic-layers/meta-mediatek-bsp/recipes-kernel/linux/linux-mtk_%.bbappend
```diff
@@ -0,0 +1,107 @@
...
+WARN_QA:remove = "buildpaths"
+
+# Torizon's required kernel settings. linux-torizon.inc adds torizon.cfg to
+# SRC_URI only for the sl1680/imx machines, and the re-merge below reads it from
+# the unpack directory, so it has to be fetched explicitly here. The require also
+# restores kernel_do_deploy:append(), whose .kernel_scm* files become the OSTree
+# oe.kernel-source metadata.
+require recipes-kernel/linux/linux-torizon.inc
+
+SRC_URI:append:lec-mtk1200 = " file://torizon.cfg"
```
rborn-tx: `recipes-kernel/linux/linux-torizon.inc` already keeps a machine list for exactly this purpose (`bb.utils.contains_any('MACHINE', 'sl1680 imx95-19x19-verdin imx93frdm', 'file://torizon.cfg', ...)`). Adding `lec-mtk-i1200-ufs` to that list instead keeps the reason in one place.

dynamic-layers/meta-mediatek-bsp/recipes-kernel/linux/linux-mtk_%.bbappend
```diff
@@ -0,0 +1,107 @@
...
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
```
rborn-tx: This resolves to `dynamic-layers/meta-mediatek-bsp/recipes-kernel/linux/linux-mtk/`, which the series never creates -- `torizon.cfg` is found through `linux-torizon.inc`'s own `FILESEXTRAPATHS`. Harmless, but dead.

conf/machine/include/common-mtk.inc
```diff
@@ -0,0 +1,70 @@
...
+
+# --- OP-TEE userspace -------------------------------------------------------
+# tee-supplicant; not pulled in by the BSP, so install it explicitly.
+IMAGE_INSTALL:append = " optee-client"
+
+# --- DRAM calibration (libdram) ---------------------------------------------
+# Without the MediaTek NDA, TF-A links prebuilt libdram, which ships only a
+# "genio-1200-evk" config (no "-ufs"). Use it for the prebuilt path, keep the
+# -ufs name for the NDA source build. This is ADLINK's own vendor default for
+# the module (meta-adlink-mtk's lec-mtk-i1200-ufs.conf sets the same).
```
rborn-tx: The parenthetical is wrong. The vendor sets `genio-1200-evk-ufs` (`meta-adlink-mtk`'s `lec-mtk-i1200-ufs.conf`), whereas this override uses `genio-1200-evk` unless `nda-mtk` is in `DISTRO_FEATURES` -- i.e. it deliberately differs from the vendor default on the default path, which is the whole point of the override. The code looks right; the comment says the opposite.

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -12,6 +12,13 @@ SRC_URI:append = " \
 
 SRC_URI:append:genericx86-64 = " file://0001-rules-whitelist-hd-devices.patch"
 
+# With the kernel cmdline's "quiet" dropped (lec-mtk-i1200-ufs.inc) to keep
+# full boot-time console output, lower console_loglevel back down via
+# sysctl.d once systemd-sysctl runs early in boot, to suppress kernel
+# messages after boot on the console/login shell - without touching the
+# boot cmdline itself. The messages remain visible via dmesg.
+SRC_URI:append:lec-mtk1200 = " file://99-printk-console.conf"
```
rborn-tx, line 20: The commit that adds this (`e012c22`, *disable Plymouth splash and route console to fbcon*) has an empty message body, in a series where every other commit explains itself well -- and it changes a recipe shared by every machine in the layer. Worth a sentence on why the splash goes away and why the loglevel is restored through `sysctl.d` rather than on the kernel cmdline.

rborn-tx:
> The change to the tmpfiles.conf is not problematic on it's own I suppose. I just find it strange that the purpose of the change is to assist in an "unofficial" method.
>
> @rborn-tx Perhaps you have some thoughts on this point.

I too think the case solved by this addition to `tmpfiles.conf` is odd. However, having an entry in that file that enforces the proper permissions to the `/var/sota` directory (independent of how it was created) looks like a useful protection to have. So, I'd be okay to accept that addition.

rborn-tx (changes requested):
@vdskvortsov I've run the whole PR past Claude, filtered out the findings considered not relevant and I'm posting what was left here for your assessment.

docs/README-genio.md
```diff
@@ -0,0 +1,315 @@
...
+and splash edits are not (TCB rejects them on raw/WIC images). Requires a
+`torizoncore-builder` release with `--raw-sector-size` support (the Genio
+system image is 4Kn) and `android-sdk-libsparse-utils` on the host.
+
+Host prerequisites
+------
+`simg2img`/`img2simg` (`android-sdk-libsparse-utils`) for `genio2img`/
+`img2genio` themselves, and a `torizoncore-builder` release with
+`--raw-sector-size` support to customize between them — either its
+`torizon/torizoncore-builder` container image, or a native install.
```
rborn-tx: The tool is only distributed as a container image - no native install.

docs/README-genio.md
rborn-tx: Users who properly install the tool would simply run:

```
torizoncore-builder build --file tcbuild-genio.yaml
```

The `docker run` invocation happens under the hood when the `torizoncore-builder` alias or function is expanded. Given that, I'd document the simpler command line above.

rborn-tx (changes requested):
Some additional comments about the docs.

vdskvortsov:
Thanks for the review, and for clearing the earlier flag.

The two converter commits are squashed as you asked — the series is now 11
commits, head `184fd061`. Everything below is in that push.

**Applied as suggested**

* `fw_env.config` no longer uses `/dev/sdb` — see the first note below for the
  form it ended up taking.
* `linux-mtk_%.bbappend`: the dead `FILESEXTRAPATHS` line is gone, and
  `torizon.cfg` moved into `linux-torizon.inc`'s existing machine list.
* `common-mtk.inc`: the `LIBDRAM_BOARD_NAME` comment was wrong, as you said —
  it now records that the non-NDA path deliberately differs from the vendor
  default.
* `README.md`: MediaTek / ADLINK LEC-MTK-i1200 row added to the machine table.
* `README-genio.md`: the `bitbake` step is dropped (the image build already
  deploys the tarball, exactly as you described); the "or a native install"
  claim is gone; step 2 is `torizoncore-builder build --file tcbuild-genio.yaml`
  rather than a raw `docker run`. While there we also dropped several passages
  that describe general `torizoncore-builder` behaviour rather than anything
  specific to this board — they belong in the TorizonCore Builder documentation,
  not in a machine README.
* The recipe, its directory and the deployed tarball are renamed to
  `genio2img-tools`; `HOMEPAGE` points at this repository; the tarball is now
  reproducible (`--sort=name --mtime=@$SOURCE_DATE_EPOCH --owner=0 --group=0`).
* The splash/console commit has a message body explaining why the splash is
  disabled and why the loglevel is restored through `sysctl.d` rather than the
  kernel command line.

**Two where the result differs from the suggestion**

*1. `fw_env.config` — a `by-path` name cannot be used here.*

We tried it first, and it fails on the board with `Cannot initialize
environment`: `libubootenv` treats `:` as the separator for a UBI volume name
and truncates the device path there, so a SCSI `by-path` node — which always
contains colons — can't be used. Verified by swapping only the device string in
an otherwise identical file: `by-path` fails, `/dev/sdb` works.

`by-id` is not an alternative either, and would be worse than the status quo:
this UFS device reports no per-LU serial, so the boot LU and the rootfs LU both
generate `scsi-1MICRON`, and it currently resolves to the **rootfs** LU.

So the series takes your other suggestion: a udev rule shipped from the
`libubootenv` bbappend gives `/dev/ufs-boot1`.

*2. Kernel fragments — the clobber is removed, but not by re-scheduling.*

Re-scheduling `do_copy_defconfig` to the `absolute-vision` position doesn't
preserve the fragments; it discards the board defconfig instead.
`KBUILD_DEFCONFIG = "defconfig"` (`linux-mtk-common.inc:72`) names the *generic*
arm64 defconfig, and `do_kernel_configme` re-runs `do_kernel_metadata` itself,
so the merge base would be rebuilt from the generic file after the task ran.

What works is clearing `KBUILD_DEFCONFIG` for this machine.
`adlink_lec_i1200_defconfig` is already in `SRC_URI` and already a merge-base
candidate — `do_kernel_metadata` drops it every build with
`defconfig was supplied both via KBUILD_DEFCONFIG and SRC_URI` — so clearing the
variable promotes it to configme's base, the fragments merge onto it normally,
and `do_copy_defconfig` has nothing left to do and is disabled. Your point about
composefs is fixed by this too: `composefs.scc` and `composefs-fsverity.scc` now
survive with no special casing.

The MediaTek SoC and evaluation-kit fragments are kept out of the merge via
`SRC_URI:remove`, and the commit message says why. In short: none of them has
ever reached this kernel — the clobber discarded them long before this change —
and each was reviewed individually against the built configuration.

**CodeRabbit's round**

Four of its findings are fixed in the same push: the GPT resize now preserves the
partition's unique GUID and type code across the delete/recreate (it already kept
the start sector and the attribute flags); `img2genio` names the directory to
`cd` into instead of `<dir>`; and the `repo` download in `README-genio.md` is
HTTPS — the same line is in `README-nvidia.md:12` if you want it swept there too.

Two I did not take. `UBOOT_BOOT_PARTITION_NUMBER = "${distro_bootpart}"` is
deliberate: bitbake has no variable of that name, leaves the reference untouched,
and U-Boot expands it at runtime — I have extended the comment there to say so.
And `$$` in the command blocks is the existing convention for "inside the crops
container", matching `README-nvidia.md`; changing it here alone would break the
convention rather than fix anything.

**One thing worth flagging**

The `torizoncore-builder` step in `README-genio.md` needs the 4Kn raw-image
support from torizon/torizoncore-builder#156. That merged on 5 August but no
published `torizon/torizoncore-builder` tag carries it yet, so the documented
flow does not work with a released tool today. Is there a numbered release
planned that would pick it up?

dynamic-layers/meta-adlink-mtk/recipes-bsp/u-boot/files/lec-mtk-i1200-ufs/fw_env.config
vdskvortsov: `by-path` turned out not to work here: `libubootenv` treats `:` as the separator
for a UBI volume name and truncates the device path there, so a SCSI `by-path`
node can't be used. `by-id` is worse — this UFS reports no per-LU serial, so the
boot LU and the rootfs LU generate the same name, and it currently resolves to
the rootfs LU.

Took the udev-rule half of your suggestion instead: `61-ufs-boot1.rules` gives
`/dev/ufs-boot1`, shipped from the `libubootenv` bbappend beside the config it
serves. More detail in the summary comment.

dynamic-layers/meta-mediatek-bsp/recipes-kernel/linux/linux-mtk_%.bbappend
vdskvortsov: Re-scheduling `do_copy_defconfig` doesn't preserve the fragments — it discards
the board defconfig instead. `KBUILD_DEFCONFIG` names the generic arm64
defconfig and `do_kernel_configme` re-runs `do_kernel_metadata` itself, so the
merge base gets rebuilt from the generic file after the task has run.

Clearing `KBUILD_DEFCONFIG` for this machine does what you were after:
`adlink_lec_i1200_defconfig` is already in `SRC_URI` and already a merge-base
candidate, so it becomes configme's base, the fragments merge onto it normally,
and `do_copy_defconfig` is left with nothing to do. The re-merge is gone, and
`composefs.scc` / `composefs-fsverity.scc` now survive with no special casing.

dynamic-layers/meta-mediatek-bsp/recipes-kernel/linux/linux-mtk_%.bbappend
vdskvortsov: Done — the machine is in that list now. One note, since the reason differs from
the other entries: with the clobber removed, `torizon.scc` merges the fragment
through the normal path, so the raw `.cfg` is needed for the post-configure
read-back in the bbappend rather than for the merge itself. The comment above
the list says so.

vdskvortsov:
@rborn-tx — your review is answered and pushed: the squash you asked for, and
the 15 inline comments. There's a summary in the comment above. Two came out
differently from what you suggested — `fw_env.config` and the kernel fragments —
and both have a reply in their own thread explaining why.

Ready for a new round of review.

conf/machine/include/common-mtk.inc
```diff
@@ -0,0 +1,69 @@
...
+DEPENDS:append:pn-trusted-firmware-a = " openssl-native"
+DEPENDS:append:pn-bl2 = " openssl-native"
+
+# Bootable partition is resolved at runtime by the distro scan (rootfs is marked
+# --active in the wks, and is not partition #1). Thread ${distro_bootpart} into
+# boot.scr and otaroot instead of hardcoding the index.
+UBOOT_BOOT_PARTITION_NUMBER = "${distro_bootpart}"
+# fitImage conf node is conf-<KERNEL_DEVICETREE, '/'->'_'>; point fdtfile/fdt_file at it.
+FITIMAGE_CONF_NODE = "${@d.getVar('KERNEL_DEVICETREE').split()[0].replace('/', '_')}"
+UENV_EXTRA_CONFIGS = "env set otaroot ${distro_bootpart} ; env set fdtfile ${FITIMAGE_CONF_NODE} ; env set fdt_file ${FITIMAGE_CONF_NODE}"
```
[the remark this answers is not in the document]
rborn-tx: Low-risk -- accepted.

rborn-tx:
@vdskvortsov Regarding:

> The torizoncore-builder step in README-genio.md needs the 4Kn raw-image support from https://github.com/torizon/torizoncore-builder/pull/156. That merged on 5 August but no published torizon/torizoncore-builder tag carries it yet, so the documented flow does not work with a released tool today. Is there a numbered release planned that would pick it up?

This addition would be available only in the early-access version of TorizonCore Builder for now. Unfortunately, an upstream problem with a Python package broke our early-access pipeline and because of that our weekly early-access releases didn't come out the past two weeks. But it should soon be fixed :pray:.

For now, it's best to omit version information from the documentation.

rborn-tx (approved):
Thanks for the contribution!

rborn-tx:
@EdTheBearded @jsrc27 @Freireg @MatheusRodrigues-tor Are you going to do a pass here?

jsrc27 (approved):

Freireg (approved):

vladimirkhusainov-emcraft:
@rborn-tx — #570 has three approvals (yours and `jsrc27`'s on 27 August, `Freireg`'s the same day), all review threads are resolved, and GitHub reports it mergeable against `scarthgap-7.x.y`.

Could it be merged?

rborn-tx:
> @rborn-tx — #570 has three approvals (yours and `jsrc27`'s on 27 August, `Freireg`'s the same day), all review threads are resolved, and GitHub reports it mergeable against `scarthgap-7.x.y`.
> 
> Could it be merged?

@vladimirkhusainov-emcraft Unfortunately not yet, but this should happen in the next few days. Per our process, we are missing the review by @EdTheBearded who is a maintainer.

EdTheBearded (approved):

EdTheBearded:
Hi @vladimirkhusainov-emcraft, sorry for the delay. The MR is finally merged :grin: 

Could you also port this to our master branch?
Please do note that we changed a bit our master branch, we split it to have all the BSP related customizations into a separated branch.
Here's the links:
[meta-torizon](https://github.com/torizon/meta-torizon)
[meta-torizon-bsp](https://github.com/torizon/meta-torizon-bsp)

vdskvortsov:
@EdTheBearded Thanks for merging. Ported, split to match the new layout:

- torizon/meta-torizon#5 — the two board-independent commits (the `resize-helper` 4Kn rootfs grow and the `/var/sota` tmpfiles entry)
- torizon/meta-torizon-bsp#7 — the rest of the port

PR #587, opened 2026-07-30
Drop HAB/AHAB status patches moved to meta-toradex-security

sergioprado:
The patches adjusting the exit code of the `hab_status` and `ahab_status` commands were carried here in `u-boot-fuse.inc`, next to a `TODO` suggesting they should live in `meta-toradex-security` instead, so that users other than Torizon OS could benefit from them.

They have now been moved to that layer, which also gained a patch for the ELE based implementation used by i.MX9x. Remove them from here, otherwise a build carrying both layers would apply the same patches twice and fail in `do_patch`.

The exit code of these commands is what makes the fusing support in `uEnv.txt` refuse to close a device that still reports HAB/AHAB events. That support is only enabled on `tdx-signed` builds, through `FUSE_SUPPORT_SIGNED_BUILD`, which is exactly when `meta-toradex-security` applies the patches, so the behavior of the fusing flow is unchanged.


sergioprado:
Obs.: This should be merged together with https://github.com/toradex/meta-toradex-security/pull/187.

jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

EdTheBearded (approved):

rborn-tx:
Guys, since this is approved by the maintainers, I'll go ahead and merge it followed by a merge on the referenced `meta-toradex-security` PR to avoid issues with the next OS build.

PR #590, opened 2026-08-06
TOR-4508 — split meta-toradex-torizon into a distro layer and a BSP layer

Freireg:
This PR is one half of a coordinated pair and **must land together with its counterpart**:

- `torizon/meta-toradex-torizon` → branch `tor-4508-distro-split`
- `torizon/meta-torizon-bsp` → branch `tor-4508-bsp-split`

Neither branch parses on its own.

### What the split does

`meta-torizon` keeps only board-agnostic distro logic. Everything keyed to a MACHINE, SoC, or vendor-BSP family moves to `meta-torizon-bsp`:

- **moves** — `verdin-* apalis-* colibri-* aquila-* toradex-smarc-* toradex-osm-imx93 lino-imx93 qemuarm64 genericx86-64 intel-* imx-generic-bsp mx8-nxp-bsp mx93-nxp-bsp mx95-nxp-bsp ti-soc am62xx beagley-ai tdx tegra imx …`
- **stays** — distro and feature overrides: `:tdx-signed :cfs-signed :cfs-support :sota :torizon-distro :common-torizon-distro :experimental`, and plain assignments

### Layer identity after the split

| | collection | priority | depends on |
|---|---|---|---|
| `meta-torizon` | `meta-torizon` | 90 | `sota virtualization-layer` |
| `meta-torizon-bsp` | `meta-torizon-bsp` | 91 | `meta-torizon` |

The BSP sits at the higher priority so its `.bbappend`s and machine configs cleanly override the distro. Both layers previously shared the collection name `meta-toradex-torizon`; that is now split, which is why they can coexist in `bblayers.conf`.

### How the commits are organized

Both branches are 4 commits, and the middle three correspond one-to-one:

| # | `meta-torizon` | `meta-torizon-bsp` |
|---|---|---|
| 1 | **naive split** — delete files that are exclusively BSP content; they live on in the BSP layer | — |
| 2 | **semantic split** — remove machine-specific *fragments* from files that stay | **semantic split** — add those same fragments here |
| 3 | **layer.conf updates** — `BBFILE_COLLECTIONS` → `meta-torizon` | **layer.conf updates** — `BBFILE_COLLECTIONS` → `meta-torizon-bsp`, priority 91, `LAYERDEPENDS = meta-torizon`, plus the global `require` of `torizon-machine-overrides.inc` |
| 4 | **documentation sources** — README points at the unified Torizon OS docs page | **documentation sources** — same |
| 5 | — | **scripts paths and templates** — `scripts/setup-environment` and `scripts/lib` follow the BSP |

The BSP branch has no naive-split commit: its equivalent was the `git filter-branch` that created the repository, so the branch begins at the semantic split.

Commits 1 and 2 on the distro side are deliberately separate. Commit 1 is pure file deletion and is easy to verify by eye; commit 2 is line-level surgery inside files that remain, and is where the real review effort belongs. **The two semantic-split commits are exact complements — every line removed by the distro's commit 2 is added by the BSP's commit 1.**

### Three mechanisms used in the semantic split

1. **Snippet in a recipe** (`.bb` / `.bbappend` / recipe `.inc`) → deleted from the distro recipe, recreated in a sibling `.bbappend` at the mirrored path in the BSP, along with any per-board data files.
2. **Snippet in a class** (classes cannot be `.bbappend`ed) **or in `torizon-base.inc`** → moved to the shared include `recipes-images/images/torizon-image-machine.inc`, pulled into each Torizon image by a one-line `.bbappend`.
3. **Snippet in distro config** (`conf/distro/include/*.inc`) → moved to `conf/machine/include/torizon-machine-overrides.inc`, `require`d globally from the BSP's `layer.conf`. This is a deliberate stopgap: many of these target Toradex modules whose machine `.conf` lives in the external Toradex BSP, so there is no per-machine `.conf` here to put them in yet. The long-term home is each machine's own `.conf`.

## Summary by CodeRabbit

- **Documentation**
  - Replaced repository-specific setup and hardware guides with links to shared Torizon OS documentation.
  - Updated contribution guidance and the project overview.

- **Refactor**
  - Renamed the layer to `meta-torizon` and simplified its configuration.
  - Removed legacy environment setup scripts and templates.

- **Chores**
  - Removed obsolete board integrations, image layouts, bootloader customizations, kernel configurations, firmware packages, and device rules.
  - Retired legacy OTA, rollback, secure-boot, flashing, and hardware-specific tooling.


Freireg:
I'll keep this as a draft while I run some more tests, but feel free to start reviewing it.

Freireg:
Promoting this from draft.

Built successfully with this layer split on **verdin-imx8mm**, **verdin-am62** and **aquila-am69**.

I also tested **beagley-ai** to validate `common-torizon` stability . A few build failures came up there, but they're all pre-existing config gaps or upstream `master` drift and reproduce independently of this MR; none relate to the split itself. That's why I'm moving this out of draft.

EdTheBearded (dismissed):

jsrc27 (dismissed):
LGTM

recipes-core/base-files/base-files_%.bbappend
```diff
@@ -19,19 +14,8 @@ do_install_basefilesissue () {
...
-do_install:append:intel-x86-common() {
-	# Replace fstab with our own for Intel Core i7-64
-	install -m 644 ${WORKDIR}/x86/fstab ${D}${sysconfdir}/fstab
-}
-
 do_install:append:cfs-support () {
 	# Get rid of the /dev/root entry in fstab to avoid errors from
 	# systemd-remount-fs.
 	sed -i -e '\#^ */dev/root#d' ${D}${sysconfdir}/fstab
 }
```
rborn-tx: ### Splitting `base-files_%.bbappend` reorders `do_install` appends

On `master` the single bbappend had three appends in this file order:

```
do_install:append:intel-x86-common()   # install our own /etc/fstab
do_install:append:cfs-support()        # sed the /dev/root line out of /etc/fstab
do_install:append:beagley-ai()         # install our own /etc/fstab
```

BitBake concatenates `:append` bodies in parse order, so on **intel-x86-common + cfs-support** the sequence was *replace, then strip `/dev/root`* — correct. (On beagley-ai + cfs-support it was *strip, then replace*, i.e. already latently broken, since `beagley-ai/fstab` also contains a `/dev/root` line.)

After the split, `cfs-support` stays in the distro bbappend and the other two move to the BSP bbappend. Because appends are ordered by layer priority (90 before 91), the new sequence is:

```
cfs-support (sed)  →  intel-x86-common (replace)  →  beagley-ai (replace)
```

So **`intel-x86-common` + `cfs-support` now ships an `/etc/fstab` that still contains `/dev/root`**, which is exactly what the sed exists to prevent (`systemd-remount-fs` errors). Both moved fstabs contain the line — I checked:

```
$ head -1 recipes-core/base-files/base-files/x86/fstab
/dev/root            /                    auto       defaults              1  1
```

Fix options, in order of preference:

1. Drop the `/dev/root` line from `x86/fstab` and `beagley-ai/fstab` outright...
2. Keep the cfs-support sed in the *BSP* bbappend, after the two replacements.
3. Convert the two replacements from `do_install:append:<machine>` to a `do_install[postfuncs]` ordered explicitly.

**Suggestion (by @rborn-tx):** make the cleanup a separate post-function, enabled only for cfs-support:

```
remove_dev_root_from_fstab() {
    sed -i -e '\#^ */dev/root#d' ${D}${sysconfdir}/fstab
}

do_install[postfuncs]:append:cfs-support = " remove_dev_root_from_fstab"
```

conf/distro/include/torizon.inc
```diff
@@ -22,18 +22,6 @@ BBMASK += " \
...
-    linux-firmware-rtl8723 \
-    linux-firmware-rtl8821 \
-    linux-firmware-rtl8822 \
-"
-
-# machines where Torizon is experimental
-EXPERIMENTALOVERRIDES = ""
-# Currently Torizon does not support any experimental machines, in future such machines can be added like below:
-#EXPERIMENTALOVERRIDES:colibri-imx6ull = ":experimental"
-DISTROOVERRIDES .= "${EXPERIMENTALOVERRIDES}"
```
rborn-tx, line 35: ### `EXPERIMENTALOVERRIDES` is deleted from `torizon.inc` and not re-added anywhere

Commit 2 removes from `conf/distro/include/torizon.inc`:

```
# machines where Torizon is experimental
EXPERIMENTALOVERRIDES = ""
# Currently Torizon does not support any experimental machines, in future such machines can be added like below:
#EXPERIMENTALOVERRIDES:colibri-imx6ull = ":experimental"
DISTROOVERRIDES .= "${EXPERIMENTALOVERRIDES}"
```

These 5 lines are the **only** semantic content in the whole diff that does not reappear in the BSP, so anyone verifying the "exact complement" claim will trip over them. It is in fact safe: `conf/distro/include/base-distro.inc:53-57` carries a byte-identical copy, and `torizon.inc:2` does `require base-distro.inc`, so the block was duplicated and the removal de-duplicates it. It also quietly fixes a latent double-`DISTROOVERRIDES .=` (harmless today only because the value is empty).

Two asks (but see the suggestion below):

- Call this out in the commit message. As written, commit 2 says "remove all the machine-specific fragments" and this is not a machine-specific fragment — it is a de-dup, and it is the one line-level claim in the PR body that a careful reviewer will find contradicted.
- The de-dup is partial. `torizon.inc` and `base-distro.inc` still both carry `PREFERRED_PROVIDER_aktualizr*` (5 lines), `SKIP_META_SECURITY_SANITY_CHECK`, `BBMASK += "/meta-freescale/recipes-graphics/cairo"` and a `TOS_SECBOOT_REQUIRED_BOOTARGS ?=` that shadows the `_DEFAULT`/`_COMMON` chain. Either finish the de-dup in a separate commit or leave all of it alone — doing one and not the others in a commit whose stated purpose is something else makes the diff harder to audit than it needs to be.

**Suggestion (by @rborn-tx):** Move the existing de-dup from the commit to a separate one; finish the de-dup on that separate commit.

classes/image_type_torizon.bbclass
```diff
@@ -80,39 +70,6 @@ TEZI_IMAGE_TEZIIMG_PREFUNCS:append = " add_signing_files_to_tezi_artifacts adjus
...
-
-# disable for now while we investigate build issues
-UBOOT_BINARY_OTA_IGNORE:aquila-am69 = "1"
-UBOOT_BINARY_OTA_IGNORE:aquila-imx95 = "1"
-UBOOT_BINARY_OTA_IGNORE:toradex-smarc-imx8mp = "1"
-UBOOT_BINARY_OTA_IGNORE:toradex-smarc-imx95 = "1"
-UBOOT_BINARY_OTA_IGNORE:verdin-imx95 = "1"
-UBOOT_BINARY_OTA_IGNORE:genericx86-64 = "1"
-UBOOT_BINARY_OTA_IGNORE:lino-imx93 = "1"
-UBOOT_BINARY_OTA_IGNORE:toradex-osm-imx93 = "1"
```
rborn-tx, line 114: ### Scope narrowing: distro-wide image settings become six-recipe settings

Four blocks moved out of *distro-scope* into `recipes-images/images/torizon-image-machine.inc`, which is pulled in by six `.bbappend`s:

| Moved from | Previously applied to |
|---|---|
| `classes/image_type_torizon.bbclass` — `UBOOT_BINARY_OTA*`, `UBOOT_BINARY_OTA_IGNORE*`, `TCB_SIGNING_*:verdin-imx8m{m,p}` | **every** image, via `IMAGE_CLASSES:append = " image_type_torizon"` in `torizon.inc:4` |
| `classes/torizon_base_image_type.inc` — `EXTRA_IMAGECMD:ota-ext4:qemuarm64` | same |
| `recipes-images/images/torizon-base.inc` — the `CORE_IMAGE_BASE_INSTALL:append:<soc>` firmware/driver block | every recipe doing `require torizon-base.inc` |
| `torizon-base.inc` — `TORADEX_PRODUCT_IDS:remove:colibri-imx6` | same |

After this PR they apply to exactly `torizon-{docker,minimal,podman}{,-dev}`. Anyone with a derived image — a customer image doing `require torizon-docker.bb` under a different recipe name, or an internal image we add later — silently loses their per-machine `UBOOT_BINARY_OTA`, the TCB signing filelist, the GPU/NPU driver installs, and the qemuarm64 ext4 tweak. `UBOOT_BINARY_OTA` going missing is the nasty one: it produces a *successfully built* OTA image with the wrong (or no) bootloader artifact.

The PR body justifies this with "classes cannot be `.bbappend`ed", which is true but not the constraint that matters. The BSP layer can ship its **own** class and inject it globally:

```
# meta-torizon-bsp/classes/torizon-image-machine.bbclass   (content of the .inc)
# meta-torizon-bsp/conf/machine/include/torizon-machine-overrides.inc
IMAGE_CLASSES:append = " torizon-image-machine"
```

That restores the original scope exactly, drops six `.bbappend` files, and removes the "did you remember to add a bbappend for your new image?" footgun. (It does interact with the BSP's layer.conf-scoping problem — see the BSP review, finding 1 — so ideally both are fixed together by hanging the include off a distro-scoped hook rather than off `layer.conf`.)

**Suggestion (by @rborn-tx):** as per "Best" suggestion in finding 1 of the BSP-layer review, the distro layer would have a `require conf/machine/include/torizon-machine-overrides.inc` which is not perfect, but simplifies everything; I'd follow that approach.

recipes-images/images/torizon-base.inc
```diff
@@ -102,53 +102,6 @@ CORE_IMAGE_BASE_INSTALL:append:sota = " \
     tzn-mqtt \
 "
 
-# Packages that install firmwares/kernel modules that must be present
-# in the rootfs. Firmware blobs, generally speaking, cannot be loaded
-# from containers as their kernel module counterparts generally expect
-# the files to be in an specific directory (generally /lib/firmware)
-
-CORE_IMAGE_BASE_INSTALL:append:mx8-nxp-bsp = " \
```
rborn-tx, line 110: ### The `:tdx` rule was applied unevenly

The PR body lists `tdx` in the **moves** column. It moved in one place and not the other:

- `recipes-core/packagegroups/packagegroup-base.bbappend` — `RDEPENDS:packagegroup-base-wifi:append:tdx` → **moved to the BSP** ✅
- `recipes-images/images/torizon-base.inc:64` — `CORE_IMAGE_BASE_INSTALL:append:tdx = " apparmor alsa-ucm-conf set-hostname tdx-info udev-toradex-rules "` → **stayed in the distro** ✗

`tdx` is a `MACHINEOVERRIDE` contributed by `meta-toradex-bsp-common`, and two of the five packages (`tdx-info`, `udev-toradex-rules`) are Toradex-BSP recipes whose `.bbappend`s now live in the BSP layer. By the PR's own rule this belongs in `torizon-image-machine.inc`. Either move it or amend the rule in the PR body to say why `:tdx` is treated differently in image recipes.

conf/distro/include/torizon.inc
```diff
@@ -22,18 +22,6 @@ BBMASK += " \
     /meta-freescale/recipes-graphics/cairo \
```
rborn-tx: ### Two vendor-BSP `BBMASK`s left behind in the distro

- `conf/distro/include/torizon.inc:23` — `BBMASK += "/meta-toradex-bsp-common/recipes-core/systemd/systemd_%.bbappend"`
- `conf/distro/include/torizon.inc:24` and `base-distro.inc:50` — `BBMASK += "/meta-freescale/recipes-graphics/cairo"`

Both are keyed to a specific vendor BSP layer, which is precisely the criterion used to move everything else. They are also the kind of thing that becomes wrong silently when the BSP layer is absent.

recipes-core/initramfs-framework/initramfs-graphics.inc
```diff
@@ -3,74 +3,6 @@
 # in the same order they are added to this variable!
 INITRAMFS_EXTRA_KMODS ?= ""
```
rborn-tx: ### `initramfs-graphics.inc` is left as an orphaned husk

After the split the file is:

```
# List of modules to be added to initramfs image.
# Important: the order here matters, since the modules will be probed
# in the same order they are added to this variable!
INITRAMFS_EXTRA_KMODS ?= ""

def get_initramfs_kmods(d):
    ...
```

The ordering comment now documents a list that lives in a different repository (and the BSP duplicated the comment there, correctly). Two nits: the leading comment should be reworded to say where the list actually is, and a 9-line file `require`d from exactly one bbappend is arguably not worth being a file any more — consider folding it into `initramfs-framework_1.0.bbappend`.

README.md
```diff
@@ -1,71 +1,32 @@
-meta-toradex-torizon
-====================
+[![Documentation](https://img.shields.io/badge/docs-online-blue)](https://torizon.github.io//)
```
rborn-tx: ### Documentation: real content deleted, replaced by a URL with a typo

The docs site is real — `https://torizon.github.io/` serves an mkdocs-material site titled "Unified documentation for the Torizon OS Yocto layers: meta-torizon (distro) and meta-torizon-bsp (BSP adaptations)", and `/contributing/` and `/common-torizon/README-ti/` resolve. So the premise is sound. But:

- **Every link is written with a double slash**: `https://torizon.github.io//` (README badge and README body) and `https://torizon.github.io//contributing/` (CONTRIBUTING). Six occurrences across the two repos. The site canonicalises to a single slash — this looks like an unsubstituted `{{ base_url }}/`. Fix before merge; it is the first thing anyone sees.
- `docs/CONTRIBUTING.md` loses the DCO explanation and the developercertificate.org link. We *require* `Signed-off-by` on every commit; the one-line "sign off your commits (`git commit -s`)" that replaces it does not tell a first-time external contributor what they are certifying. It also drops the `scope: brief description` commit-message convention that this very PR follows.
- It also drops "Open a second pull request to the `master` branch, unless not applicable" while `.github/workflows/backport.yml` and the `backport-to-*` labels are still in place. Contributors now have no in-repo description of the two-branch workflow.
- `docs/CONTRIBUTING.md` ends without a trailing newline (`\ No newline at end of file`).
- The README no longer says where to report issues. That section was repo policy, not build documentation, and "open an issue here or a topic on the community forum" does not obviously belong on a docs site shared by two layers.
- Deleting `docs/README-*.md` and `docs/setup-environment.md` from the distro is right — they follow the BSP. But see the BSP review (finding 5): those files are now stale in their new home, and the distro README's pointer to `docs/setup-environment.md` was deleted rather than repointed at the BSP.

README.md
```diff
@@ -1,71 +1,32 @@
...
-See [docs/setup-environment.md](./docs/setup-environment.md) for how the build-environment setup scripts work and how to add support for a new vendor or board.
+Documentation
+=============
+All build guides and project documentation are unified on the Torizon OS docs
+site:
 
-Development Process
-===================
-Torizon is maintained by the Toradex R&D team. Development happens in this repository, including issues, PRs, and discussions. This repository is then used by our internal CI/CD infrastructure.
+### 📖 https://torizon.github.io//
```
rborn-tx: ### Documentation reorganization

@Freireg Was the reorganization of the documentation and the creation of a separate repository for its maintenance discussed?

In my opinion, doing so is taking on another responsibility (maintaining a separate documentation repo) with the extra risk of having documentation that is not sync'ed with the implementation. Given that, I'd like to understand what gains we envision with this reorg and whether alternatives have been considered (e.g. using the developer website).

@EdTheBearded @jsrc27 Opinions on this?

rborn-tx (commented):
Made a review with Claude, dropped irrelevant findings and added my comments/questions/suggestions on top.

README.md
jsrc27: Honestly, I do vaguely remember the documentation in `torizon.github.io` being brought up prior. Cause this PR isn't the first time I've seen it. That said I don't recall any in-depth discussion about this specifically.

At the moment I don't have strong opinions about it. All the various Toradex/Torizon maintained repos do their own thing regarding documentation I suppose. I wonder if this should be streamlined, but that's probably a bigger conversation now.

README.md
Freireg: @rborn-tx, fixed it!

recipes-core/base-files/base-files_%.bbappend
Freireg: @rborn-tx I think suggestion 2: 
> Keep the cfs-support sed in the BSP bbappend, after the two replacements.

is most likely the safest one here that keeps the layer interaction as close as it is currently. But can it pass as something that should be present in the BSP layer? I can't see how this would be a distro-specific append.

conf/distro/include/torizon.inc
Freireg: @rborn-tx this section delete was reverted and the de-duplication is on the latest commit

conf/distro/include/torizon.inc
Freireg: @rborn-tx removed in the semantic split commit.

recipes-images/images/torizon-base.inc
Freireg: @rborn-tx good observation, would that also apply to `:tdx-signed` right? I kept those in the distro as well.

rborn-tx: I tend to agree but I'd be okay with leaving the few occurrences still on `meta-torizon` for now, since their placement is somewhat arguable.

Freireg: @rborn-tx done!

recipes-core/initramfs-framework/initramfs-graphics.inc
Freireg: @rborn-tx done!

classes/image_type_torizon.bbclass
Freireg: @rborn-tx done! I'll resolve the comment on the other PR as well

README.md
Freireg: What @jsrc27 said is correct: this idea was briefly mentioned during the early stages of planning the layer split and was later discussed further among me, @EdTheBearded, and @MatheusRodrigues-tor. I should have included both of you in those discussions, and I apologize for that oversight.

The goal is to provide a single home for documentation shared by `meta-torizon` and `meta-torizon-bsp`. Because the two layers are designed to be used together, duplicating build instructions and contribution guidance across both repositories would increase the maintenance burden and create more opportunities for the documentation to diverge.

MatheusRodrigues-tor (dismissed):

recipes-core/base-files/base-files_%.bbappend
rborn-tx: Notice options 1..3 are Claude's proposals. Mine is indicated separately and I believe is the one that best keeps previous behavior (rather: it even fixes a latent bug on beagley-ai as mentioned by Claude).

Now if the question is where it should live, then I'd agree that it fits better in the BSP adaptation layer.

README.md
rborn-tx: If that's the goal, I'd suggest simply referencing one layer from the other (in whatever direction we see most sensible):

- For build instructions, probably having them in the BSP layer and referencing from the distro is more sensible.
- For the contributions guide, we could do the same or even duplicate it since they are short, very rarely updated, and they would likely have small differences between the two layers.

Doing that would avoid the burden that you mentioned plus the issues that I mentioned: extra burden for maintaining yet another repository and the possibility of having code/metadata not synchronized with documentation.

NOTE: And there is also the possibility of having the documentation in the developer website, as I mentioned before, though I personally see not as a great option due to the additional time it may take for docs changes to be accepted and go online.

Anyways, we should hear more opinions... Cc @EdTheBearded @griloHBG

conf/distro/include/torizon.inc
rborn-tx: Ok, I'll consider this solved here and add a related comment on the other PR.

classes/image_type_torizon.bbclass
rborn-tx: @Freireg Could you please mention our decision of accepting the dependency `meta-torizon` -> `conf/machine/include/torizon-machine-overrides.inc` (provided by `meta-torizon-bsp`) in the commit `layer-split: semantic split on distro side` message body.

recipes-core/base-files/base-files_%.bbappend
Freireg: Right, I moved it to the bsp layer, but  the suggested `do_install[postfuncs]:append:cfs-support` syntax caused a parse error. So I changed the implementation slightly:
``` 
remove_dev_root_from_fstab() {
  # Get rid of the /dev/root entry in fstab to avoid errors from
  # systemd-remount-fs.
  sed -i -e '\#^ */dev/root#d' ${D}${sysconfdir}/fstab
}

BASE_FILES_POSTFUNCS = ""
BASE_FILES_POSTFUNCS:cfs-support = "remove_dev_root_from_fstab"

do_install[postfuncs] += " ${BASE_FILES_POSTFUNCS}"

```
Should have the same result with the `sed` beign applied if the cfs-support is enabled and only after all the installs had ocurred.

This change has already been pushed to the BSP layer as part of the semantic split. If you’d like to continue discussing the cleanup implementation, I’d prefer that we continue the discussion there.

classes/image_type_torizon.bbclass
rborn-tx: Sorry for the nitpick, but the commit message looks like a meeting note :grin: 

How about:

```
Accept meta-torizon dependency on file torizon-machine-overrides.inc,
to be normally provided by meta-torizon-bsp.

Require this file from base-distro.inc to scope machine and SoC
overrides to Torizon distros while preserving their original distro-wide
application.
```

README.md
griloHBG: IMO most (if not all) of the documentation available at https://torizon.github.io/ must be available at https://developer.toradex.com. In the github page, we see things that are expected to be found in Toradex Developer:

- layer architecture and migration guide
- how to build Torizon OS
- guides for non-Toradex devices

Not having these information at Toradex Developer would make people confused (mainly users/customers).

As for Contributing/Development guide, where to report issues and  licensing information, these information could simply be made available in the `README.md`, or in their respective `CONTRIBUTING.md` and `LICENSE` files - whatever feels better. 

We can think about keeping both, but I feel like doing so will just duplicate the effort without much gain from my POV.

But I am open to other POVs.

classes/image_type_torizon.bbclass
Freireg: @rborn-tx done!

README.md
rborn-tx: Ok, that mostly aligns with my view.

@EdTheBearded Could we settle on not having the new documentation repo for now. Or do you have a strong opinion for having it?

EdTheBearded: Right, sorry! I forgot to answer here.
No, no strong opinions here... My initial idea for this suggestion was to have a single source of truth between the 2 new repo, so we try and avoid duplicating info in the READMEs.
But after our discussing @rborn-tx, I kind of agree that it might be too much work, having yet another repo to maintain.
So I'm completely fine ditching the new doc and keeping everything in the READMEs.

Freireg: Ok, I'll move all the build instructions to the BSP while keeping the contributions guide shared between both (since we probably will not change this one that much).

Freireg: Are you guys ok with me adding the "Achitecture & Migration" section on the layer's documentation? @rborn-tx @EdTheBearded

conf/layer.conf
```diff
@@ -4,39 +4,9 @@ BBPATH .= ":${LAYERDIR}"
...
-    tegra:${LAYERDIR}/dynamic-layers/meta-tegra/*/*/*.bb \
-    tegra:${LAYERDIR}/dynamic-layers/meta-tegra/*/*/*.bbappend \
-    toradex-bsp-common-layer:${LAYERDIR}/dynamic-layers/meta-toradex-bsp-common/*/*/*.bb \
-    toradex-bsp-common-layer:${LAYERDIR}/dynamic-layers/meta-toradex-bsp-common/*/*/*.bbappend \
-    toradex-ti-layer:${LAYERDIR}/dynamic-layers/meta-toradex-ti/*/*/*.bb \
-    toradex-ti-layer:${LAYERDIR}/dynamic-layers/meta-toradex-ti/*/*/*.bbappend \
-    virtualization-layer:${LAYERDIR}/dynamic-layers/meta-virtualization/*/*/*.bb \
-    virtualization-layer:${LAYERDIR}/dynamic-layers/meta-virtualization/*/*/*.bbappend \
-"
+LAYERDEPENDS_meta-torizon = "sota virtualization-layer"
```
Freireg, line 11: @rborn-tx since we created the BSP dependency in `torizon.inc` and `common-torizon.inc` with the `require .../bsp-machine-overrides.inc`, shouldn't we explicitly declare the `meta-torizon-bsp` in the distro's `LAYERDEPENDS`?

README.md
rborn-tx: Having a dedicated architecture document sounds sensible to me.

The [document](https://github.com/torizon/torizon.github.io/blob/master/docs/architecture-migration.md) currently on the site seems focused on the transition (layer split).

I'd reshape its contents to focus on the architecture **as it is now**, likely dropping background on how it came to be.

conf/layer.conf
EdTheBearded: I think it makes sense. Though it's not a hard dependency, without it we won't build anything.

MatheusRodrigues-tor (dismissed):

conf/layer.conf
rborn-tx: Hmm, I don't think so. Two reasons:

- As I mentioned in the other thread, we are not depending on the layer, we are depending on a file `bsp-machine-overrides.inc` being provided by some other layer.
- This would represent a circular dependency since `meta-torizon-bsp` has:
  ```
  11:LAYERDEPENDS_meta-torizon-bsp = "meta-torizon"
  ```

README.md
rborn-tx: I see you've added the arch doc to `meta-torizon-bsp`. My impression is that an architecture doc would belong here, but that's not a strong opinion. If you prefer to keep most docs there, then I'd suggest mentioning the arch doc from the README in this repo as well.

Freireg: @rborn-tx  that sounds reasonable. I’d prefer to keep the architecture document in meta-torizon-bsp, since that is where the BSP and build-integration documentation now lives, but linking to it from this repository will make it easier to discover, so I’ll add that reference.

rborn-tx (approved):
Latest version looks good to me.

EdTheBearded (approved):

MatheusRodrigues-tor (approved):

jsrc27 (approved):

PR #591, opened 2026-08-07
docker-moby: fix silent drop of per-device block-I/O options

dmitrys-emcraft:
The Docker client shipped by meta-virtualization on this branch (Docker
25.0.3) is affected by docker/cli #5321:
`ThrottledeviceOpt.GetList()` allocates its result slice with length 0 and
capacity `len(values)`, so the `copy()` into it copies nothing.
`--device-read-bps`, `--device-write-bps`, `--device-read-iops` and
`--device-write-iops` are parsed and validated, then handed to the daemon
as an empty list and silently dropped, with no error, no warning, and no
limit applied to the container. The kernel capability underneath is
unaffected; `--blkio-weight-device` takes a different code path and is not
affected.

Affects every board on this branch's meta-virtualization pin (`f92518e2`),
not just one platform - the pin predates the affected range's exit
(fixed upstream in v25.0.7 and v27.1.2).

Commit:

- docker-moby: fix silent drop of per-device block-I/O options
  Backport docker/cli `73e78a582222` as a source patch, instead of
  advancing `SRCREV_cli`, which would pull in 104 commits across 183 files.
  Drop this patch once the meta-virtualization revision advances past
  `f92518e2`.

Validation: confirmed the four options are silently dropped on unpatched
25.0.3 and correctly applied to the container's block-I/O cgroup with
this patch, on the ADLINK LEC-MTK-i1200 (MediaTek Genio 1200).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Docker device throttling configuration so configured throttle devices are correctly returned and applied instead of appearing as an empty list.

### DIFF
--- a/recipes-containers/docker/docker-moby/0002-opts-fix-ThrottledeviceOpt-GetList-returning-an-empty.patch
+++ b/recipes-containers/docker/docker-moby/0002-opts-fix-ThrottledeviceOpt-GetList-returning-an-empty.patch
@@ -1,0 +1,50 @@
+From 73e78a5822224bd7640888b6b5c2ab6b3f35bd13 Mon Sep 17 00:00:00 2001
+From: Jianyong Wu <wujianyong@hygon.cn>
+Date: Wed, 7 Aug 2024 10:27:21 +0800
+Subject: [PATCH] run: fix GetList return empty issue for throttledevice
+
+Test "--device-read-bps" "--device-write-bps" will fail. The root
+cause is that GetList helper return empty as its local variable
+initialized to zero size.
+
+This patch fix it by setting the related slice size to non-zero.
+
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):

EdTheBearded (approved):

EdTheBearded:
Checked and Wrynose `docker-moby` is at v29.3.0. So no need to port it to master.

rborn-tx (commented):
LGTM

@dmitrys-emcraft Thanks for your contribution!

Freireg (approved):
LGTM

EdTheBearded:
@MatheusRodrigues-tor any comments here? Can we merge this one?

rborn-tx:
Since maintainers already approved it, I'll go ahead and merge this.

PR #593, opened 2026-08-11
torizon-conf: replace per-container tags with a single major based one

EdTheBearded:
Every CT_TAG_<NAME> variable in containers-tags.sh was hardcoded based on the current container versioning. This made sense in the past, since we had different release tags for different container, but now that it's all unified, we don't need individual tags anymore. We collapse them into one CT_TAG and substitute it from ${TDX_MAJOR}, now having only one tag for all containers, and matching the OS major version.

Related-to: TOR-4526

## Summary by CodeRabbit

* **Enhancements**
  * Standardized container image tag configuration across the system.
  * Container tags are now automatically aligned with the installed platform release version.
  * Improved consistency when selecting and running platform-provided container images.
  * Reduced maintenance overhead by using a unified tag configuration.
  * Supports more reliable container image selection across supported platform components.


recipes-core/torizon-conf/torizon-conf_1.0.bb
```diff
@@ -20,4 +20,5 @@ do_install () {
     install -m 0755 ${UNPACKDIR}/machine.sh ${D}${sysconfdir}/profile.d/
     sed -i "s/@@MACHINE@@/${MACHINE}/g" ${D}${sysconfdir}/profile.d/machine.sh
     install -m 0755 ${UNPACKDIR}/containers-tags.sh ${D}${sysconfdir}/profile.d/
+    sed -i "s/@@TAG@@/${TDS_MAJOR}/g" ${D}${sysconfdir}/profile.d/containers-tags.sh
```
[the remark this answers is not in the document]
EdTheBearded: Oops, corrected.
Since I tested this on scarthgap then copied the change to master, this could explain the typo here.

recipes-core/torizon-conf/torizon-conf/containers-tags.sh
```diff
@@ -1,21 +1,4 @@
...
-export CT_TAG_QT6_WAYLAND="4"
-export CT_TAG_QT6_WAYLAND_EXAMPLES="4"
-export CT_TAG_RT_TESTS="4"
-export CT_TAG_STRESS_TESTS="4"
-export CT_TAG_WAYLAND_BASE="4"
-export CT_TAG_WAYLAND_GTK_3="4"
-export CT_TAG_WESTON="4"
-export CT_TAG_WESTON_TOUCH_CALIBRATOR="4"
+export CT_TAG="@@TAG@@"
+
```
rborn-tx: Is the extra empty line here intended?

LeoBelmont:
Hi guys! I want to add to the discussion some information that could be useful.

The TCCP team decided to match the tags with the OS version, but also have an alias for the debian version and yocto branch. So for example an image for TOS8 will have tag 8 and also forky-wrynose. This is what you already expect.

For non LTS yocto branches or Torizon releases, such as Walnascar, we would have the tag (for example) trixie-walnascar, and no numbered tag since Torizon doesn't have a major for these. What we did for imx95 Walnascar build was release also under tag 4, which was technically incorrect even though it works. In the future we would like for these to have their own proper tags instead of mixing with something different.

I imagine that you could just use the same sed logic for these non LTS builds with the string instead of number tag, but mentioning here in case its useful.

recipes-core/torizon-conf/torizon-conf/containers-tags.sh
EdTheBearded: Nope, removed!

EdTheBearded:
> Hi guys! I want to add to the discussion some information that could be useful.
> 
> The TCCP team decided to match the tags with the OS version, but also have an alias for the debian version and yocto branch. So for example an image for TOS8 will have tag 8 and also forky-wrynose. This is what you already expect.
> 
> For non LTS yocto branches or Torizon releases, such as Walnascar, we would have the tag (for example) trixie-walnascar, and no numbered tag since Torizon doesn't have a major for these. What we did for imx95 Walnascar build was release also under tag 4, which was technically incorrect even though it works. In the future we would like for these to have their own proper tags instead of mixing with something different.
> 
> I imagine that you could just use the same sed logic for these non LTS builds with the string instead of number tag, but mentioning here in case its useful.

Not sure what you meant here.
Having a tag linked to major is enough? Would you like custom suffix for the tag, with a variable to control? Or a second tag that's more specific than the major (like one tag just the major number, the other the major + suffix)?

LeoBelmont:
To summarize what I was saying:

We will use tags that match the OS version for LTS builds (tag 8 for Torizon 8, tag 9 for Torizon 9 etc). So these changes in the PR are enough for Torizon LTS majors.

If we need to make containers for a non LTS Torizon OS (which outputs version 0.0.0 right?) we can't make a number tag, so we will only use a string like forky-blacksail for example.

recipes-core/torizon-conf/torizon-conf/containers-tags.sh
```diff
@@ -1,21 +1,3 @@
...
-export CT_TAG_QT5_WAYLAND_EXAMPLES="4"
-export CT_TAG_QT6_WAYLAND="4"
-export CT_TAG_QT6_WAYLAND_EXAMPLES="4"
-export CT_TAG_RT_TESTS="4"
-export CT_TAG_STRESS_TESTS="4"
-export CT_TAG_WAYLAND_BASE="4"
-export CT_TAG_WAYLAND_GTK_3="4"
-export CT_TAG_WESTON="4"
-export CT_TAG_WESTON_TOUCH_CALIBRATOR="4"
+export CT_TAG="@@TAG@@"
```
EdTheBearded, line 3: I've created this to hack github and have an organized thread. Just ignore the code fragment above... :grinning: 

I was talking to @LeoBelmont offline, and we're discussing the tag format, for the "walnascar" special case.

If we follow the format `<major>`, if we ever face a "walnascar" scenario again, we would have a tag collision. Out of that there are 2 possibilities:
1. @LeoBelmont suggested that we use tags of the format `<debian version>-<yocto version>`. e.g. `forky-blacksail`
2. I suggested we could use `<torizon major>-<yocto version>`. e.g. `8-wrynose`

Both would be sufficient to avoid the collision scenario, but on option 1, we'd need to manually update the recipe with version changes. Where on option 2, the tags would be completely hands-off on Torizon side.
We'd of course prefer option 2, since it's zero maintenance, but I'd like to hear @LeoBelmont, João Fracarolli and Vitória Bastos opinions.

LeoBelmont:
After discussing with the teams, we reached the conclusion that we will maintain 3 tags:

1. `<yocto-branch>` tag, for example `wrynose`
2. `<debian-version>-<yocto-branch>` tag, for example `forky-wrynose`
3. `<torizon-major>` tag, always matching the Torizon major. For wrynose the tag would be `8`. These numbered tags will only exist for Torizon LTS releases.

If Torizon only ever plans to have LTS releases, it's safe to use the numbered tags (like this PR already does), otherwise I recommend switching to Option 1 I listed here (yocto branch) as it will always work even for non LTS releases.

Freireg (approved):

recipes-core/torizon-conf/torizon-conf/containers-tags.sh
rborn-tx: From @LeoBelmont's [comment](https://github.com/torizon/meta-toradex-torizon/pull/593#issuecomment-5410322958): to handle non-LTS (such as `walnascar`) we'd need to set the `CT_TAG` to the yocto branch name here.

Does this unblock this thread?

EdTheBearded: Yes, but due to the layer split, I've opened this new MR
https://github.com/torizon/meta-torizon/pull/6

EdTheBearded:
Closed this PR, since now the layer is split and we need to move this to another repository.
Link: https://github.com/torizon/meta-torizon/pull/6

PR #602, opened 2026-09-04
Export TCB re-signing tarball for the verdin-am62 and refactoring of the tarball generation

rborn-tx:
TorizonCore Builder re-signs a pre-built image's bootloader with the customer's own keys; on the Verdin AM62 the artifacts it needs never leave the U-Boot work directory. This branch exports them as a tarball in the Tezi image, as the iMX8M already does, and first restructures that mechanism, which cannot describe a two-multiconfig machine.

## Main changes

- **`recipes-bsp/tcb-signing-files/`** (new, TOR-4583): the tarball moves out of a `do_image_teziimg` prefunc that tarred a glob list out of `DEPLOY_DIR_IMAGE` into `WORKDIR`. It is now built once by a recipe of its own, shared through sstate, buildable without an image, and packed reproducibly (sorted members, `SOURCE_DATE_EPOCH` mtimes, `0/0` ownership). Fixes a stale tarball shipped with signing off, and per-image repacking of the deploy directory. Building it in a configuration that has signing off now fails with a message naming the reason, and it is kept out of `bitbake world`.
- **`image_type_torizon.bbclass`**: the image side is now a conditional dependency plus a conditional `TEZI_ARTIFACTS` entry. `TCB_SIGNING_SUPPORT` is hashed by its computed value, so a signature moves when a machine switches the export on or off, not when a helper is reworded.
- **`u-boot-tcb-sign.inc`** (TOR-4543): the AM62 export. The two multiconfigs produce colliding basenames (`u-boot.dtb`, `spl/*`, everything under `arch/arm/dts/`) with different content, so instead of a flat file set the export is a tree with one binman working directory per multiconfig (`a53/`, `k3r5/`), 41 files: unpack, enter a root, run binman. It also deploys a key-free `u-boot.dtb` snapshot, taken before oe-core injects the build's FIT key, and a `manifest.json` giving format version, machine, U-Boot revision and expected roots. No key material is exported. Gated on `TCB_SIGNING_SUPPORT`, keyed on `TDX_K3_SECBOOT_ENABLE`, plus a second form for the FIT-less k3r5 multiconfig.

iMX8M tarball name, contents and placement unchanged.

## Migration

Existing build directories need one cleanup before the first build, and the recipe to clean differs by machine. Fresh directories, and builds served from a shared sstate cache, are unaffected.

- **`verdin-imx8mp` / `verdin-imx8mm`** — the tarball's owner moves from the image recipe to the new one, and sstate refuses to install over the file already sitting in `DEPLOY_DIR_IMAGE`: `Recipe tcb-signing-files is trying to install files into a shared area when those files already exist`. Run `bitbake -c cleansstate <image>` once; the tarball then repacks to identical contents.
- **`verdin-am62`** — the pristine-DTB guard fires once on a tree that already holds a signed U-Boot. This branch moves `do_uboot_assemble_fitimage`'s signature and not `do_compile`'s, so BitBake re-runs the former alone over the `u-boot.dtb` the previous build signed in place. Run `bitbake -c cleansstate u-boot-toradex-ti` once. The same guard fires after `UBOOT_SIGN_KEYDIR` or `UBOOT_SIGN_KEYNAME` changes, where it also stops a key-free file being exported with the previous key still in it.

## Tests

- All five AM62 artifacts re-signed from the tarball alone; payloads byte-identical to the deployed binaries (certificates differ — the checking toolchain predates the companion `meta-toradex-security` pinning).
- iMX8M: twelve members and archive bytes identical.
- Tarball built standalone, reproducible across repacks, in the Tezi image once and unmodified.
- Feature off: nothing built, stale tarball not shipped.
- Each multiconfig deploys only its own root.
- The feature gate exercised both ways on `verdin-imx8mp`: with signing on `do_compile` runs and packs the twelve members; with `TCB_SIGNING_SUPPORT` forced to `0` the build stops at that task with the recipe's own error.
- The migration above checked against the signatures rather than assumed: `do_compile`'s signature data for `u-boot-toradex-ti` contains no reference to anything this branch adds, while `do_uboot_assemble_fitimage`'s does.

Related-to: TOR-4543  
Related-to: TOR-4583

## Summary by CodeRabbit

- **New Features**
  - Added support for packaging bootloader files required to re-sign Torizon OS images.
  - Added Trusted Computing Base signing support for supported Verdin i.MX8 and AM62 platforms.
  - Added AM62 signing exports for A53 and K3R5 targets, including versioned manifests.
  - Torizon OS image builds now include the required signing package when signing support is enabled.

- **Improvements**
  - Improved U-Boot artifact handling for machine-specific configurations and signing workflows.
  - Added validation for missing signing artifacts and unsupported configuration selections.

### DIFF
--- a/recipes-bsp/tcb-signing-files/tcb-signing-files.bb
+++ b/recipes-bsp/tcb-signing-files/tcb-signing-files.bb
@@ -1,0 +1,76 @@
+SUMMARY = "Bootloader artifacts needed by TorizonCore Builder to re-sign an image"
+DESCRIPTION = "Packs the intermediate, pre-signing bootloader artifacts that \
+TorizonCore Builder needs to recreate and re-sign the bootloader of a pre-built \
+Torizon OS image. This recipe aggregates output deployed by other recipes; it \
+builds nothing of its own."
+SECTION = "BSP"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+require recipes-bsp/tcb-signing-files/tcb-signing-files.inc
+
+inherit deploy nopackages
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+COMPATIBLE_MACHINE = "(verdin-imx8mp|verdin-imx8mm)"
+
+# bitbake world would otherwise build this on machines where the feature is off
+# and nothing ever deploys the files below.
+EXCLUDE_FROM_WORLD = "${@'0' if d.getVar('TCB_SIGNING_SUPPORT') == '1' else '1'}"
+
+INHIBIT_DEFAULT_DEPS = "1"
+
+do_fetch[noexec] = "1"
+do_unpack[noexec] = "1"
+do_patch[noexec] = "1"
+do_configure[noexec] = "1"
+do_install[noexec] = "1"
+
+B = "${WORKDIR}/build"
+
+# DEPLOY_DIR_IMAGE is otherwise just a directory that may not have been filled
+# in yet, so wait on the deploy tasks of the recipes that fill it.
+do_compile[depends] += "${@' '.join('%s:do_deploy' % r for r in d.getVar('TCB_SIGNING_INPUT_DEPENDS').split())}"
+
+# The filelist decides the archive's contents, and the enable condition decides
+# whether anything downstream uses it.
+do_compile[vardeps] += "TCB_SIGNING_FILELIST TCB_SIGNING_SUPPORT"
+
+do_compile() {
+    # The inputs are deployed only when the feature is on; without this the tar
+    # below fails on its own with a bare "Cannot stat".
+    if [ "${TCB_SIGNING_SUPPORT}" != "1" ]; then
+        bbfatal "TCB_SIGNING_SUPPORT is not \"1\" for MACHINE \"${MACHINE}\", so" \
+                "nothing deploys the files this recipe packs. Build it only in a" \
+                "configuration that enables the signing feature."
+    fi
+
+    if [ -z "${TCB_SIGNING_FILELIST}" ]; then
+        bbfatal "TCB_SIGNING_FILELIST is empty for MACHINE \"${MACHINE}\", so there" \
+                "is nothing to pack. Either define it or keep this recipe out of the" \
+                "build (see TCB_SIGNING_SUPPORT)."
+    fi
+
+    # Without the flags below, member mtimes, ownership and order track the
+    # build, so the same sources would not produce the same archive twice.
+    tar_timestamp_args=""
+    if [ -n "${SOURCE_DATE_EPOCH}" ]; then
+        tar_timestamp_args="--mtime=@${SOURCE_DATE_EPOCH} --clamp-mtime"
+    fi
+
+    # The shell expands the filelist's globs where the command runs, not where
+    # tar reads. --format is pinned so a new tar default cannot move the bytes.
+    (cd "${DEPLOY_DIR_IMAGE}" && tar \
+        --dereference --hard-dereference \
+        --sort=name --format=gnu \
+        ${tar_timestamp_args} \
+        --owner=0 --group=0 --numeric-owner \
+        -czf "${B}/${TCB_SIGNING_FILES_TARBALL}" ${TCB_SIGNING_FILELIST})
+}
+do_compile[dirs] = "${B}"
+do_compile[cleandirs] = "${B}"
+
+do_deploy() {
+    install -m 0644 "${B}/${TCB_SIGNING_FILES_TARBALL}" "${DEPLOYDIR}/${TCB_SIGNING_FILES_TARBALL}"
+}
+addtask deploy after do_compile before do_build
--- a/recipes-bsp/tcb-signing-files/tcb-signing-files.bb
+++ b/recipes-bsp/tcb-signing-files/tcb-signing-files.bb
@@ -12,7 +12,9 @@ require recipes-bsp/tcb-signing-files/tcb-signing-files.inc
 inherit deploy nopackages
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
-COMPATIBLE_MACHINE = "(verdin-imx8mp|verdin-imx8mm)"
+# Anchored: unanchored, "verdin-am62" would also match verdin-am62p and the
+# k3r5 multiconfig's verdin-am62-k3r5, neither of which this recipe packs for.
+COMPATIBLE_MACHINE = "^(verdin-imx8mp|verdin-imx8mm|verdin-am62)$"
 
 # bitbake world would otherwise build this on machines where the feature is off
 # and nothing ever deploys the files below.
@@ -31,6 +33,10 @@ B = "${WORKDIR}/build"
 # DEPLOY_DIR_IMAGE is otherwise just a directory that may not have been filled
 # in yet, so wait on the deploy tasks of the recipes that fill it.
 do_compile[depends] += "${@' '.join('%s:do_deploy' % r for r in d.getVar('TCB_SIGNING_INPUT_DEPENDS').split())}"
+
+# Some machines build part of their input in another multiconfig, which needs an
+# edge of its own; see TCB_SIGNING_INPUT_MCDEPENDS.
+do_compile[mcdepends] += "${TCB_SIGNING_INPUT_MCDEPENDS}"
 
 # The filelist decides the archive's contents, and the enable condition decides
 # whether anything downstream uses it.
--- a/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
+++ b/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
@@ -1,0 +1,35 @@
+# Definitions shared by both sides of the "TCB signing files" tarball: the
+# intermediate, pre-signing bootloader artifacts that TorizonCore Builder needs
+# to re-sign a pre-built Torizon OS image. The tcb-signing-files recipe builds
+# and deploys it; classes/image_type_torizon.bbclass puts it in the Tezi image.
+
+# Name of the tarball, as deployed and as seen inside the Tezi image.
+TCB_SIGNING_FILES_TARBALL = "tcb_signing_files.tar.gz"
+
+TCB_SIGNING_FILES_RECIPE = "tcb-signing-files"
+
+def is_hab_signed_bootloader_and_fit_enabled(d):
+    if d.getVar('TDX_IMX_HAB_ENABLE') == '1' and d.getVar('UBOOT_SIGN_ENABLE') == '1':
+        return '1'
+
+    return '0'
+
+# Whether this build exports the TCB signing files at all. What the condition
+# reads comes from the globally inherited signing class, so a recipe and an
+# image resolve it the same way.
+TCB_SIGNING_SUPPORT ?= "0"
+TCB_SIGNING_SUPPORT:verdin-imx8mp ?= "${@is_hab_signed_bootloader_and_fit_enabled(d)}"
+TCB_SIGNING_SUPPORT:verdin-imx8mm ?= "${@is_hab_signed_bootloader_and_fit_enabled(d)}"
+
+# What goes into the tarball, relative to DEPLOY_DIR_IMAGE. The shell expands
+# these in that directory, so globs work and an entry matching nothing fails
+# the build rather than producing a short tarball.
+TCB_SIGNING_FILELIST ?= ""
+TCB_SIGNING_FILELIST:verdin-imx8mp ?= "uboot_config bl31* lpddr4_pmu_train_* u-boot.dtb u-boot-nodtb.bin spl/ u-boot-dtbs/"
+TCB_SIGNING_FILELIST:verdin-imx8mm ?= "uboot_config bl31* lpddr4_pmu_train_* u-boot.dtb u-boot-nodtb.bin spl/ u-boot-dtbs/"
+
+# The do_deploy tasks that put those files in DEPLOY_DIR_IMAGE. Keep this in
+# step with TCB_SIGNING_FILELIST: a file packed but not depended upon is a race.
+TCB_SIGNING_INPUT_DEPENDS ?= ""
+TCB_SIGNING_INPUT_DEPENDS:verdin-imx8mp ?= "virtual/bootloader ${IMX_DEFAULT_ATF_PROVIDER} ${IMX_EXTRA_FIRMWARE}"
+TCB_SIGNING_INPUT_DEPENDS:verdin-imx8mm ?= "virtual/bootloader ${IMX_DEFAULT_ATF_PROVIDER} ${IMX_EXTRA_FIRMWARE}"
--- a/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
+++ b/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
@@ -20,9 +20,8 @@ def is_hab_signed_bootloader_and_fit_enabled(d):
 
     return '0'
 
-# The TI K3 counterpart of the above, kept separate: TCB_SIGNING_SUPPORT is in
-# the packing task's vardeps, so rewording the iMX helper would repack the
-# tarball on machines that ship today.
+# The TI K3 counterpart of the above. A separate function so that each names the
+# condition it tests, and the k3r5 multiconfig needs a third form anyway.
 def is_k3_secboot_and_fit_enabled(d):
     if d.getVar('TDX_K3_SECBOOT_ENABLE') == '1' and d.getVar('UBOOT_SIGN_ENABLE') == '1':
         return '1'
@@ -47,6 +46,11 @@ TCB_SIGNING_SUPPORT:verdin-imx8mp ?= "${@is_hab_signed_bootloader_and_fit_enable
 TCB_SIGNING_SUPPORT:verdin-imx8mm ?= "${@is_hab_signed_bootloader_and_fit_enabled(d)}"
 TCB_SIGNING_SUPPORT:verdin-am62 ?= "${@is_k3_secboot_and_fit_enabled(d)}"
 TCB_SIGNING_SUPPORT:verdin-am62-k3r5 ?= "${@is_k3_secboot_enabled(d)}"
+
+# Hash by the value this resolves to: otherwise every task reading it also
+# depends on the helper's text, which BitBake captures down to the comments
+# that follow it.
+TCB_SIGNING_SUPPORT[vardepvalue] = "${TCB_SIGNING_SUPPORT}"
 
 # What goes into the tarball, relative to DEPLOY_DIR_IMAGE. The shell expands
 # these in that directory, so globs work and an entry matching nothing fails
--- a/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
+++ b/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
@@ -1,12 +1,18 @@
-# Definitions shared by both sides of the "TCB signing files" tarball: the
-# intermediate, pre-signing bootloader artifacts that TorizonCore Builder needs
-# to re-sign a pre-built Torizon OS image. The tcb-signing-files recipe builds
-# and deploys it; classes/image_type_torizon.bbclass puts it in the Tezi image.
+# Definitions shared by everything that handles the "TCB signing files" tarball:
+# the intermediate, pre-signing bootloader artifacts that TorizonCore Builder
+# needs to re-sign a pre-built Torizon OS image. The tcb-signing-files recipe
+# builds and deploys it, image_type_torizon.bbclass puts it in the Tezi image,
+# and u-boot-tcb-sign.inc deploys what it packs.
 
 # Name of the tarball, as deployed and as seen inside the Tezi image.
 TCB_SIGNING_FILES_TARBALL = "tcb_signing_files.tar.gz"
 
 TCB_SIGNING_FILES_RECIPE = "tcb-signing-files"
+
+# Root of the export tree, for machines that export a tree instead of loose
+# files. The U-Boot recipe installs into it and the file list packs it, so the
+# name lives here rather than being repeated on both sides.
+TCB_SIGNING_EXPORT_DIR ?= "tcb-signing"
 
 def is_hab_signed_bootloader_and_fit_enabled(d):
     if d.getVar('TDX_IMX_HAB_ENABLE') == '1' and d.getVar('UBOOT_SIGN_ENABLE') == '1':
--- a/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
+++ b/recipes-bsp/tcb-signing-files/tcb-signing-files.inc
@@ -20,12 +20,33 @@ def is_hab_signed_bootloader_and_fit_enabled(d):
 
     return '0'
 
+# The TI K3 counterpart of the above, kept separate: TCB_SIGNING_SUPPORT is in
+# the packing task's vardeps, so rewording the iMX helper would repack the
+# tarball on machines that ship today.
+def is_k3_secboot_and_fit_enabled(d):
+    if d.getVar('TDX_K3_SECBOOT_ENABLE') == '1' and d.getVar('UBOOT_SIGN_ENABLE') == '1':
+        return '1'
+
+    return '0'
+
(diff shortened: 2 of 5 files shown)


rborn-tx:
@lucas-akira Tagging you here to let you know about the refactoring of the re-signing tarball generation.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -48,6 +96,292 @@ do_deploy:append:verdin-imx8mm() {
...
+        bbfatal "\"${UBOOT_DTB_BINARY}\" not found in \"$(pwd)\"; it is required" \
+                "to export a pristine DTB for re-signing (and without it the" \
+                "build could not embed its own FIT key either)."
+    fi
+
+    if ! fdtget -l "${UBOOT_DTB_BINARY}" / >/dev/null 2>&1; then
+        bbfatal "\"$(pwd)/${UBOOT_DTB_BINARY}\" is not a readable device tree blob."
+    fi
+
+    if fdtget -l "${UBOOT_DTB_BINARY}" /signature >/dev/null 2>&1; then
```
rborn-tx, line 124: **The pristine-DTB guard fails the first incremental build after this lands**

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:124`

```sh
    if fdtget -l "${UBOOT_DTB_BINARY}" /signature >/dev/null 2>&1; then
        bbfatal "\"$(pwd)/${UBOOT_DTB_BINARY}\" already contains a /signature node," \
                "so it is not a pristine DTB. This usually means" \
                "do_uboot_assemble_fitimage() ran again without do_compile();" \
                "clean the recipe and rebuild."
    fi
```

The guard is right about the condition it detects. The problem is that this branch is what causes
that condition, on every build directory that already has a signed verdin-am62 U-Boot in it.

`concat_dtb:prepend:verdin-am62` (line 159) is new text inside `concat_dtb()`, which
`uboot_assemble_fitimage_helper()` calls, which `do_uboot_assemble_fitimage` calls. Adding it moves
that task's signature. Nothing in this branch touches `do_compile`, so `do_compile`'s signature
does not move and its stamp stays valid. So on the next build bitbake re-runs
`do_uboot_assemble_fitimage` alone, with `${B}/${config}/u-boot.dtb` still holding the
`/signature` node that the *previous* build's `concat_dtb()` wrote into it in place. oe-core writes
the key with `mkimage -K "${UBOOT_DTB_BINARY}"` and only afterwards copies to `u-boot.dtb-signed`
(`meta/classes-recipe/uboot-sign.bbclass:140` and `:169`), so the working-tree DTB is the signed
one from then on.

Reachability: every existing build directory where `MACHINE = verdin-am62`, `tdx-signed` is
inherited and U-Boot has been built at least once. That is precisely the configuration this feature
targets. Builds with signing off are unaffected, because the prepend's body is gated on
`TCB_SIGNING_SUPPORT`.

The same failure recurs later on any change that moves only `do_uboot_assemble_fitimage`'s inputs
and not `do_compile`'s. Switching `UBOOT_SIGN_KEYDIR` or `UBOOT_SIGN_KEYNAME` from development keys
to production keys is the obvious one, and it is a workflow people run.

The PR description's migration note (`bitbake -c cleansstate <image>`) does not help here: the
recipe that needs cleaning is `u-boot-toradex-ti`.

Fix: take the snapshot in `do_compile:append:verdin-am62` instead of in a `concat_dtb:prepend`.
`${B}/${config}/u-boot.dtb` is pristine at the end of `do_compile` (the injection happens in the
later task), so the snapshot lands at the same content, the snapshot and the DTB it copies are
produced by one task and can no longer disagree, and `do_compile` then carries
`TCB_SIGNING_SUPPORT` in its signature, so flipping the feature forces the recompile that makes the
guard's precondition true. Keep the `/signature` check as a real assertion once it can no longer
fire on ordinary rebuilds.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
```diff
@@ -0,0 +1,70 @@
...
+        --sort=name --format=gnu \
+        ${tar_timestamp_args} \
+        --owner=0 --group=0 --numeric-owner \
+        -czf "${B}/${TCB_SIGNING_FILES_TARBALL}" ${TCB_SIGNING_FILELIST})
+}
+do_compile[dirs] = "${B}"
+do_compile[cleandirs] = "${B}"
+
+do_deploy() {
+    install -m 0644 "${B}/${TCB_SIGNING_FILES_TARBALL}" "${DEPLOYDIR}/${TCB_SIGNING_FILES_TARBALL}"
```
rborn-tx, line 80: **The new recipe collides with the tarball the image used to deploy**

`recipes-bsp/tcb-signing-files/tcb-signing-files.bb:68`

```sh
do_deploy() {
    install -m 0644 "${B}/${TCB_SIGNING_FILES_TARBALL}" "${DEPLOYDIR}/${TCB_SIGNING_FILES_TARBALL}"
}
```

Before this branch, `pack_tcb_signing_binaries_in_teziimg()` copied the tarball into
`${IMGDEPLOYDIR}`, from where `do_image_complete`'s sstate installed it into `DEPLOY_DIR_IMAGE`
under the *image* recipe's manifest. The name has not changed, so on an existing build directory
`${DEPLOY_DIR_IMAGE}/tcb_signing_files.tar.gz` is already there and already owned by someone else.

sstate refuses that outright. `meta/classes-global/sstate.bbclass:289` calls `bb.fatal("Recipe %s is
trying to install files into a shared area when those files already exist. ...")` for any
`sharedfiles` entry that exists on disk. Ordering makes it unavoidable rather than a race:
`do_image_teziimg[depends]` pulls `tcb-signing-files:do_deploy` in *before* the image's
`do_image_complete` runs, so the new recipe installs first and the stale file has not yet been
cleaned by the image's own manifest cleanup.

Reachability: existing build directories for `verdin-imx8mp` or `verdin-imx8mm` with
`TDX_IMX_HAB_ENABLE` and `UBOOT_SIGN_ENABLE` both on. The AM62 never shipped a tarball, so it is
not affected. Fresh builds and shared-sstate builds are fine, since the image's task hash moves and
no pre-PR object matches it.

Fix: the PR description already names the remedy, so this is a matter of putting it where it
survives. Add the migration note to commit `ae946ae`'s message and to the release notes: an
existing build directory needs `bitbake -c cleansstate <image>` (or the file removed by hand)
before the first build. Quoting the sstate error text in the note would save the next person the
grep.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
```diff
@@ -0,0 +1,70 @@
...
+# Some machines build part of their input in another multiconfig, which needs an
+# edge of its own; see TCB_SIGNING_INPUT_MCDEPENDS.
+do_compile[mcdepends] += "${TCB_SIGNING_INPUT_MCDEPENDS}"
+
+# The filelist decides the archive's contents, and the enable condition decides
+# whether anything downstream uses it.
+do_compile[vardeps] += "TCB_SIGNING_FILELIST TCB_SIGNING_SUPPORT"
+
+do_compile() {
+    if [ -z "${TCB_SIGNING_FILELIST}" ]; then
```
rborn-tx, line 54: **`bitbake tcb-signing-files` fails whenever the feature is off**

`recipes-bsp/tcb-signing-files/tcb-signing-files.bb:42`

```sh
do_compile() {
    if [ -z "${TCB_SIGNING_FILELIST}" ]; then
        bbfatal "TCB_SIGNING_FILELIST is empty for MACHINE \"${MACHINE}\", so there" \
                "is nothing to pack. Either define it or keep this recipe out of the" \
                "build (see TCB_SIGNING_SUPPORT)."
    fi
```

The guard checks the one condition that cannot happen and skips the one that can. All three
machines in `COMPATIBLE_MACHINE` define `TCB_SIGNING_FILELIST`, so the `bbfatal` above is
unreachable. What is reachable is `TCB_SIGNING_SUPPORT = "0"`, which is the default for every
build that does not inherit `tdx-signed`. Nothing then deploys the inputs, and `tar` dies on its
own:

- verdin-am62: `tcb-signing/` does not exist, so `tar: tcb-signing: Cannot stat: No such file or
  directory`, exit 2.
- verdin-imx8m*: `u-boot.dtb` and `u-boot-nodtb.bin` are deployed only by uboot-sign's
  `deploy_dtb()`, which oe-core gates on `UBOOT_SIGN_ENABLE`
  (`meta/classes-recipe/uboot-sign.bbclass:205`, `:212`), so those two entries are missing and tar
  exits 2 the same way.

I ran both shapes of the command to confirm the exit code and the message.

Reachability: `bitbake tcb-signing-files`, which commit `ae946ae` advertises as one of the
benefits of the refactoring, and `bitbake world`, which the recipe joins because it sets no
`EXCLUDE_FROM_WORLD`. Neither goes through the `TCB_SIGNING_SUPPORT` gate that protects the image
path. Before this branch the tar only ever ran under that gate, so this is new exposure.

Fix: two lines. Replace the unreachable filelist check with a `TCB_SIGNING_SUPPORT != "1"`
`bbfatal` carrying the same advice, and add
`EXCLUDE_FROM_WORLD = "${@'0' if d.getVar('TCB_SIGNING_SUPPORT') == '1' else '1'}"` so `world`
stays green. The layer already uses `EXCLUDE_FROM_WORLD` in `recipes-core/torizon-conf` and
`recipes-core/torizon-users`.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -48,6 +96,292 @@ do_deploy:append:verdin-imx8mm() {
...
+        bbfatal "\"${srcdir}/${TCB_SIGNING_UBOOT_DTB_NOKEYS}\" not found, but FIT" \
+                "signing is enabled and this machine exports a pristine DTB for" \
+                "TCB re-signing. The concat_dtb:prepend that produces it did not" \
+                "run -- check whether oe-core's uboot-sign.bbclass still calls" \
+                "concat_dtb()."
+    fi
+
+    # Deployed under its build-directory name, without the versioned file and
+    # symlink deploy_dtb() makes, for consumers with no binman tree to read.
+    install -m 0644 "${srcdir}/${TCB_SIGNING_UBOOT_DTB_NOKEYS}" \
```
rborn-tx, line 153: **The loose `u-boot.dtb-nokeys` in the deploy directory has no consumer**

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:153`

```sh
    install -m 0644 "${srcdir}/${TCB_SIGNING_UBOOT_DTB_NOKEYS}" \
                    "${DEPLOYDIR}/${TCB_SIGNING_UBOOT_DTB_NOKEYS}"
```

Commit `6e4b747` introduced this deploy so the snapshot would be reachable outside the U-Boot
workdir, which made sense at that point in the series. By the end of the series the A53 export
takes its own copy straight from the build directory (line 225), and
`TCB_SIGNING_FILELIST:verdin-am62` packs only `${TCB_SIGNING_EXPORT_DIR}/`, so the loose file is
never packed and never read. I grepped the whole layer: `TCB_SIGNING_UBOOT_DTB_NOKEYS` appears
only in this file.

Consequence is small: one unversioned file in `DEPLOY_DIR_IMAGE`, unlike every other U-Boot deploy
artifact, which carries `-${MACHINE}-${PV}-${PR}` and a symlink.

Fix: drop `tcbsgn_deploy_pristine_dtb()`'s install and keep only its existence check (which the
A53 export needs anyway), or, if an out-of-tree consumer does read it, say which one in the
comment. The comment currently says "for consumers with no binman tree to read", which does not
name anybody.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
```diff
@@ -0,0 +1,70 @@
...
+    # Without the flags below, member mtimes, ownership and order track the
+    # build, so the same sources would not produce the same archive twice.
+    tar_timestamp_args=""
+    if [ -n "${SOURCE_DATE_EPOCH}" ]; then
+        tar_timestamp_args="--mtime=@${SOURCE_DATE_EPOCH} --clamp-mtime"
+    fi
+
+    # The shell expands the filelist's globs where the command runs, not where
+    # tar reads. --format is pinned so a new tar default cannot move the bytes.
+    (cd "${DEPLOY_DIR_IMAGE}" && tar \
```
rborn-tx, line 69: **Nothing checks that the packed tree holds the roots the manifest advertises**

`recipes-bsp/tcb-signing-files/tcb-signing-files.bb:57` and
`recipes-bsp/u-boot/u-boot-tcb-sign.inc:179`

`TCB_SIGNING_MANIFEST_ROOTS` is a fixed `"k3r5 a53"`, written by the A53 side alone, while each
root is deployed by the multiconfig that owns it under its own `TCB_SIGNING_SUPPORT` value. The
`tcb-signing-files.inc` comment at lines 31-39 already works through the case where the two
disagree and concludes it only wastes files. That holds in the direction where the k3r5 half is on
and the A53 half is off. In the other direction (`TDX_K3_SECBOOT_ENABLE:verdin-am62-k3r5 = "0"`,
say) the A53 side writes a manifest promising a `k3r5` root, `tar tcb-signing/` succeeds over
whatever is present, and the short export ships for the consumer to notice.

That reading is consistent with what the manifest is for, so this is not a bug so much as an
inconsistency with the rest of the series. Commit `4276931` makes a point of failing `do_deploy()`
rather than shipping a short export, and `tcb-signing-files.inc:56-57` says an entry matching
nothing should fail the build rather than produce a short tarball. The build knows which roots it
deployed and could apply the same rule.

Fix: in `do_compile`, for the machines that export a tree, check that every root named in
`${TCB_SIGNING_EXPORT_DIR}/manifest.json` (or in a `TCB_SIGNING_MANIFEST_ROOTS` promoted to the
shared include) is a directory before tarring, and `bbfatal` if not.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -48,6 +96,292 @@ do_deploy:append:verdin-imx8mm() {
...
+
+# One root per multiconfig inside the export tree. TCB_SIGNING_EXPORT_DIR names
+# the tree itself and is defined in the include required above.
+TCB_SIGNING_EXPORT_DIR_A53 ?= "${TCB_SIGNING_EXPORT_DIR}/a53"
+TCB_SIGNING_EXPORT_DIR_K3R5 ?= "${TCB_SIGNING_EXPORT_DIR}/k3r5"
+
+# Bump the format version only for a change consumers have to notice; the roots
+# list must name every root the deploy code produces.
+TCB_SIGNING_MANIFEST_VERSION ?= "1"
+TCB_SIGNING_MANIFEST_ROOTS ?= "k3r5 a53"
```
rborn-tx, line 179: **Root names are spelled twice**

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:173-179`

```
TCB_SIGNING_EXPORT_DIR_A53 ?= "${TCB_SIGNING_EXPORT_DIR}/a53"
TCB_SIGNING_EXPORT_DIR_K3R5 ?= "${TCB_SIGNING_EXPORT_DIR}/k3r5"
...
TCB_SIGNING_MANIFEST_ROOTS ?= "k3r5 a53"
```

`a53` and `k3r5` appear in both places, and a rename of one without the other produces a manifest
that lies about the tree. This is the same duplication commit `2f98a8f` removed for
`TCB_SIGNING_EXPORT_DIR`, so it is worth closing the same way: derive the roots list from the two
directory variables with `os.path.basename`, or define the two basenames once and build the paths
from them.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
```diff
@@ -0,0 +1,70 @@
...
+    if [ -z "${TCB_SIGNING_FILELIST}" ]; then
+        bbfatal "TCB_SIGNING_FILELIST is empty for MACHINE \"${MACHINE}\", so there" \
+                "is nothing to pack. Either define it or keep this recipe out of the" \
+                "build (see TCB_SIGNING_SUPPORT)."
+    fi
+
+    # Without the flags below, member mtimes, ownership and order track the
+    # build, so the same sources would not produce the same archive twice.
+    tar_timestamp_args=""
+    if [ -n "${SOURCE_DATE_EPOCH}" ]; then
```
rborn-tx, line 63: **Two dead branches in the packing recipe**

`recipes-bsp/tcb-signing-files/tcb-signing-files.bb:51`

```sh
    if [ -n "${SOURCE_DATE_EPOCH}" ]; then
```

`SOURCE_DATE_EPOCH` is exported unconditionally by `meta/conf/bitbake.conf:685` and falls back to a
fixed `1302044400` when no source is available, which is this recipe's case. The empty branch never
runs. Harmless, but it reads as though the variable might be unset, which would undercut the
reproducibility claim if it were true.

The unreachable `TCB_SIGNING_FILELIST` check at line 42 is the other one; finding 3 covers it.

rborn-tx (commented):
Findings by Claude for assessment.

recipes-bsp/tcb-signing-files/tcb-signing-files.inc
```diff
@@ -0,0 +1,79 @@
...
+# What goes into the tarball, relative to DEPLOY_DIR_IMAGE. The shell expands
+# these in that directory, so globs work and an entry matching nothing fails
+# the build rather than producing a short tarball.
+TCB_SIGNING_FILELIST ?= ""
+TCB_SIGNING_FILELIST:verdin-imx8mp ?= "uboot_config bl31* lpddr4_pmu_train_* u-boot.dtb u-boot-nodtb.bin spl/ u-boot-dtbs/"
+TCB_SIGNING_FILELIST:verdin-imx8mm ?= "uboot_config bl31* lpddr4_pmu_train_* u-boot.dtb u-boot-nodtb.bin spl/ u-boot-dtbs/"
+
+# The AM62 packs the tree whole: its two multiconfigs contribute artifacts with
+# the same basenames, so a glob list of loose files cannot separate them.
+TCB_SIGNING_FILELIST:verdin-am62 ?= "${TCB_SIGNING_EXPORT_DIR}/"
```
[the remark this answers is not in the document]
rborn-tx: Wrong conclusion - declined.

recipes-bsp/tcb-signing-files/tcb-signing-files.inc
```diff
@@ -0,0 +1,79 @@
...
+TCB_SIGNING_INPUT_DEPENDS:verdin-imx8mm ?= "virtual/bootloader ${IMX_DEFAULT_ATF_PROVIDER} ${IMX_EXTRA_FIRMWARE}"
+
+# One producer for the AM62, not the four its artifacts come from: the U-Boot
+# recipe installs the whole export tree out of its own workdir and sysroot.
+TCB_SIGNING_INPUT_DEPENDS:verdin-am62 ?= "virtual/bootloader"
+
+# Inputs built in another multiconfig, which TCB_SIGNING_INPUT_DEPENDS cannot
+# express: an mcdepends entry has to name the multiconfigs it crosses.
+TCB_SIGNING_INPUT_MCDEPENDS ?= ""
+TCB_SIGNING_INPUT_MCDEPENDS:verdin-am62 ?= "mc::k3r5:virtual/bootloader:do_deploy"
```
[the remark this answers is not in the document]
rborn-tx: Wrong conclusion - declined.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Declined; the guard firing here is intended. A migration note naming `bitbake -c cleansstate u-boot-toradex-ti` will go into the commit message instead.

Key rotation argues for keeping the hook where it is. Change `UBOOT_SIGN_KEYDIR` but keep the key name, and `mkimage -K` rewrites the existing key node in place: the build succeeds, and `u-boot.dtb-nokeys` ships with a key embedded in it. The guard is what stops that.

`do_compile:append` also runs in the k3r5 multiconfig, where `concat_dtb` never does, so the snapshot would rest on the machine override alone.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
rborn-tx: Accepted. `ae946ae` will carry the migration with the sstate error text quoted, so it survives the merge.

Worth adding: the AM62 needs a different cure. The pristine-DTB re-run guard fires on an existing signed AM62 tree for the same underlying reason — `do_uboot_assemble_fitimage`'s hash moves while `do_compile`'s does not — and there the recipe to clean is `u-boot-toradex-ti`, not the image. That note goes into `6e4b747`.

The PR description currently names only the image, which is right for the iMX8M and silent on the AM62.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
rborn-tx: Accepted, with one correction and one refusal.

`EXCLUDE_FROM_WORLD` and a `TCB_SIGNING_SUPPORT` guard land in `ae946ae`, before merge, so the iMX8M signature move folds into the one this branch already causes rather than adding a second.

Correction: the iMX8M `do_deploy` appends are ungated, unlike the AM62 pair. With HAB off and FIT on, every filelist entry is present and `tar` succeeds. The trigger there is `UBOOT_SIGN_ENABLE`, not `TCB_SIGNING_SUPPORT`.

Refusal: the filelist check stays. It is what fires when a machine joins `COMPATIBLE_MACHINE` without a filelist.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Declined. The deployed copy is not a leftover: the commit that produces it can be applied without the export-tree commits that follow it, and then this file is the only thing the snapshot publishes. That is the class the comment's "consumers with no binman tree to read" refers to, rather than a named consumer.

The deploy directory is also the interface a layer publishes on. An artifact reachable only inside one machine's tarball is packaged, not published — so having no in-tree reader is not by itself a reason to drop it.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
rborn-tx: Declined. A deploy-side check — the natural place, since the manifest is written there — cannot work: nothing orders the two multiconfigs' `do_deploy` tasks against each other, the only mcdepends edge being on `tcb-signing-files:do_compile`. The A53 append would race the k3r5 one.

That leaves the shared `do_compile` body, which is deliberately kept free of machine-specific logic: the iMX8M would take a signature move for a feature it does not have.

Reachability is thin too. `TDX_K3_SECBOOT_ENABLE` has no `:k3r5` override, so the two values agree unless a `local.conf` line separates them.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Declined, though it is the closest call of the set.

The `TCB_SIGNING_EXPORT_DIR` precedent is not parallel: that name was duplicated across two files and could drift unseen. These two spellings sit six lines apart, under comments that describe each other.

More decisive, neither proposed form protects the edit that could actually happen. `a53` and `k3r5` are core names fixed by the hardware, so a rename is not the risk; adding a third root is — and both forms still require editing the roots list for that.

recipes-bsp/tcb-signing-files/tcb-signing-files.bb
rborn-tx: Declined, with one correction: the `1302044400` fallback never applies here. This distro sets `SOURCE_DATE_EPOCH = "0"`, and the generated `run.do_compile` emits `--mtime=@0`.

That is precisely why the test is `-n` and not a truthiness test. `bitbake.conf` warns, directly above the fallback, that a `SOURCE_DATE_EPOCH` of `0` may be misread as no SDE — and this distro sets exactly that. `if sde:` or `[ "$sde" != "0" ]` would read a deliberate value as unset and quietly disable reproducibility. The branch stays to keep that mistake out of reach.

MatheusRodrigues-tor (approved):

jsrc27 (approved):

EdTheBearded (approved):

EdTheBearded:
@rborn-tx is this something we want to port to master branch?

Freireg (approved):

rborn-tx:
> @rborn-tx is this something we want to port to master branch?

It is, but since signed builds are mostly broken on `master` I created a separate ticket for that port (to be tackled once signed builds are properly working).

PR #605, opened 2026-09-10
u-boot: record the signing conditions in the AM62 manifest

rborn-tx:
TorizonCore Builder re-signs a pre-built image's bootloader with the customer's own keys, then checks its work by comparing bytes against the original. Whether that comparison can succeed depends on two things the AM62 signing export does not state: the epoch its certificates were dated from, and whether their validity was pinned at all. This adds them to the manifest the export already carries.

Summary of changes:

- `tcb-signing/manifest.json` gains a `signing` object with three fields.
- `source_date_epoch`: the value binman ran with, from the task environment.
- `certificates_pinned`: measured by reading `notBefore` out of a deployed certificate and comparing it with `SOURCE_DATE_EPOCH`, not derived from `TDX_K3_SECBOOT_REPRODUCIBLE`. Unreadable is `false`.
- `openssl`: the version that produced them, as provenance.
- New helper `tcbsgn_am62_certificates_pinned` takes that measurement, on a k3r5 artifact.
- `format_version` stays at 1. AM62 only; the iMX8M machines are untouched.

Related-to: TOR-4543

<details>
  <summary>Details</summary>

  ## Main changes

  `recipes-bsp/u-boot/u-boot-tcb-sign.inc` — `tcb-signing/manifest.json` gains a `signing` object:
   
  ```json
  "signing": {
    "source_date_epoch": 0,
    "certificates_pinned": true,
    "openssl": "3.5.7"
  }
  ```

  - **`source_date_epoch`** — the value binman ran with. Both sides of the re-signing path hardcode zero today, independently, with nothing connecting them; recording it turns that accidental agreement into one a consumer can check.
  - **`certificates_pinned`** — measured, not derived from `TDX_K3_SECBOOT_REPRODUCIBLE`. The signing tool sets the validity period only when its openssl is 3.4 or newer, and otherwise warns and carries on, so a build can ask for reproducible certificates and deploy wall-clock ones. The value is established by reading `notBefore` back out of a deployed certificate and comparing it with `SOURCE_DATE_EPOCH`. Unreadable is `false`, which costs a consumer nothing it was not already doing.
  - **`openssl`** — provenance, in the same spirit as `uboot.release`: a different X.509 encoder is the first suspect when bytes differ beyond the validity fields. Not something to branch on.

  The certificate read is a k3r5 one: `tiboot3-am62x-hs-fs-verdin.bin` is the only export member that begins with a certificate, `tispl.bin` and `u-boot.img` being FIT images. Both halves are signed by one source tree through one patch, so the result generalises, but the field is documented as that measurement rather than as a claim about every certificate exported.

  `format_version` stays at 1: no released image carries this manifest, so there is no v1 consumer to protect.

  Requested by TCB-898, which consumes the export. AM62 only; the iMX8M machines carry no manifest and are untouched.

  ## Tests

  On `verdin-am62` with signing enabled:
   
  - Manifest correct in the packed tarball, not only in the deploy directory: `source_date_epoch` a JSON number, `certificates_pinned` a JSON boolean, mode `0644`.
  - Built without `-f`, so task selection is itself the evidence: 1 task of 2505 re-ran, `do_deploy`. `do_compile` does not move, so no U-Boot rebuild and no signature churn elsewhere.
  - Negative case: with `TDX_K3_SECBOOT_REPRODUCIBLE = "0"` the field reads `false`, and the certificate corroborates it independently — wall-clock `notBefore`, `notAfter` exactly 30 days later, fresh random serial.
  - The comparison exercised against a non-zero epoch, since the build's is zero and zero is also what several failure modes produce: exact match `true`, one second out `false`, one day out `false`. Unaffected by timezone (host `-03`, container UTC).
  - Reproducible: after `bitbake -c cleansstate u-boot-toradex-ti mc:k3r5:u-boot-toradex-ti` and a rebuild from source on both multiconfigs, the tarball is byte-identical and the certificate serial unchanged.

## Summary by CodeRabbit

* **New Features**
  * AM62 TCB signing manifests now include signing provenance, including the source date, certificate validity pinning status, and the OpenSSL version used.
  * The source date is recorded as unavailable when no value is provided.

* **Bug Fixes**
  * Certificate pinning status is now reported as false when certificates are missing, unreadable, or contain invalid dates.

### DIFF
--- a/recipes-bsp/u-boot/u-boot-tcb-sign.inc
+++ b/recipes-bsp/u-boot/u-boot-tcb-sign.inc
@@ -11,7 +11,7 @@ require recipes-bsp/tcb-signing-files/tcb-signing-files.inc
 # two branches as oe-core's do_uboot_assemble_fitimage(). More than one
 # UBOOT_MACHINE entry is fatal, since every deploy name here is flat. The result
 # comes back in a variable because BitBake emits only the shell functions it
-# sees called, and a call inside a command substitution is not one of them.
+# sees called, and scarthgap's parser misses a call inside a quoted expansion.
 tcbsgn_uboot_set_config_dir() {
     if [ -n "${UBOOT_CONFIG}" ]; then
         set -- ${UBOOT_MACHINE}
@@ -271,14 +271,40 @@ tcbsgn_deploy_am62_export_a53() {
     done
 }
 
-# Manifest at the root of the export tree: which layout to expect, and which
-# U-Boot produced the artifacts, so that a consumer whose binman comes from
-# another revision is detectable instead of silently producing different output.
-# The A53 side writes it alone, since everything in it is build-wide and both
-# multiconfigs build one recipe at one SRCREV.
+# Echoes whether the certificate's validity was pinned. Measured, not taken
+# from the switch that asks for it: the signing tool pins validity only when
+# its openssl is new enough, so intent and outcome can differ. Pass a k3r5
+# artifact: the A53's own outputs are FIT images, whose certificates are not
+# at offset 0. Unreadable is "false", the answer that costs a consumer least.
+tcbsgn_am62_certificates_pinned() {
+    local notbefore epoch
+    local certfile="${1?certificate file}"
+
+    [ -f "${certfile}" ] || { echo "false"; return 0; }
+
+    notbefore=$(openssl x509 -inform DER -in "${certfile}" -noout -startdate \
+                2>/dev/null | sed -ne 's/^notBefore=//p')
+    [ -n "${notbefore}" ] || { echo "false"; return 0; }
+
+    # Task scripts run under set -e, where a bare assignment from a failing
+    # command ends the shell on dash even though bash survives it.
+    epoch=$(date -u -d "${notbefore}" +%s 2>/dev/null) || epoch=""
+    if [ -n "${epoch}" -a "${epoch}" = "${SOURCE_DATE_EPOCH}" ]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+}
+
+# Manifest at the root of the export tree: which layout to expect, which U-Boot
+# produced the artifacts, and under which conditions they were signed, so a
+# consumer whose binman differs is detectable instead of silently producing
+# different output. The A53 side writes it alone: everything in it is
+# build-wide, and both multiconfigs build one recipe at one SRCREV.
 tcbsgn_write_am62_manifest() {
     local outdir="${DEPLOYDIR}/${TCB_SIGNING_EXPORT_DIR}"
     local verfile ubootversion ubootrelease roots sep r
+    local pinned opensslver epoch
 
     tcbsgn_uboot_set_config_dir
     tcbsgn_uboot_set_config_type
@@ -306,10 +332,29 @@ tcbsgn_write_am62_manifest() {
         sep=", "
     done
 
+    # The a53 root's own copy, installed earlier in this same append, so this
+    # reaches into no other multiconfig's deploy directory.
+    pinned=$(tcbsgn_am62_certificates_pinned \
+             "${DEPLOYDIR}/${TCB_SIGNING_EXPORT_DIR_A53}/tiboot3-am62x-hs-fs-verdin.bin")
+
+    # Provenance only: this task's openssl, which the k3r5 side takes from the
+    # same recipe at the same version. Consumers do not branch on it.
+    opensslver=$(openssl version | awk '{print $2}')
+    if [ -z "${opensslver}" ]; then
+        bbfatal "\"openssl version\" produced nothing; the manifest records it" \
+                "as the provenance of the certificates the export carries."
+    fi
+
+    # Empty would leave the field with no value at all, which is not JSON. A
+    # shell default here would read as BitBake's and be the shell's, so: two lines.
+    epoch="${SOURCE_DATE_EPOCH}"
+    [ -n "${epoch}" ] || epoch="null"
+
     install -d "${outdir}"
-    printf '{\n  "format_version": %s,\n  "machine": "%s",\n  "uboot": {\n    "version": "%s",\n    "release": "%s"\n  },\n  "roots": [%s]\n}\n' \
+    printf '{\n  "format_version": %s,\n  "machine": "%s",\n  "uboot": {\n    "version": "%s",\n    "release": "%s"\n  },\n  "roots": [%s],\n  "signing": {\n    "source_date_epoch": %s,\n    "certificates_pinned": %s,\n    "openssl": "%s"\n  }\n}\n' \
            "${TCB_SIGNING_MANIFEST_VERSION}" "${MACHINE}" \
            "${ubootversion}" "${ubootrelease}" "${roots}" \
+           "${epoch}" "${pinned}" "${opensslver}" \
            > "${outdir}/manifest.json"
 
     # Explicit mode: the tarball preserves it, and a umask-dependent one would


recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -306,10 +330,20 @@ tcbsgn_write_am62_manifest() {
         sep=", "
     done
 
+    # The a53 root's own copy, installed earlier in this same append, so this
+    # reaches into no other multiconfig's deploy directory.
+    pinned=$(tcbsgn_am62_certificates_pinned \
```
rborn-tx, line 337: ### `certificates_pinned` is measured on the k3r5 root, but the gate it detects is evaluated per multiconfig

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:335`

```sh
    # The a53 root's own copy, installed earlier in this same append, so this
    # reaches into no other multiconfig's deploy directory.
    pinned=$(tcbsgn_am62_certificates_pinned \
             "${outdir}/a53/tiboot3-am62x-hs-fs-verdin.bin")
```

`tiboot3-am62x-hs-fs-verdin.bin` is built by the k3r5 multiconfig and staged into the A53
recipe sysroot through mcdepends, so what this reads is a k3r5 certificate. The commit message
says as much and defends it: one source tree, one patch, so the result generalises.

The part that does not follow from "one source tree" is the openssl version. In the patch, the
decision to pin is taken inside binman at signing time:

```python
        epoch = os.environ.get('SOURCE_DATE_EPOCH')
        if epoch is not None:
            if self.version_tuple() >= MIN_VALIDITY_VERSION:
```

`self.version_tuple()` is the openssl each multiconfig actually invokes, out of its own
`recipe-sysroot-native`. The two multiconfigs share a recipe and a `SRC_URI`, but they do not
share a native sysroot, so the gate is answered twice. A `PREFERRED_VERSION_openssl-native`
scoped to one multiconfig, or the two multiconfigs resolving openssl-native differently for any
other reason, gives `certificates_pinned: true` measured on k3r5 while the A53 root's own
`tispl.bin` and `u-boot.img` certificates carry wall-clock validity. A consumer that byte
compares would then fail on exactly the files the manifest encouraged it to trust.

Reachability is low. It needs a deliberately asymmetric openssl-native, which nobody has today.
The documentation gap is what makes this worth raising: every caveat lives in the commit message
and in the comment above the helper, and the consumer reads neither. `manifest.json` is a
contract with a tool in another repository (TCB-898), and the layer has no schema document
anywhere, so a bare `certificates_pinned: true` is all the consumer gets.

Fix, cheapest first. Say in the manifest what the field measures, rather than leaving the reader
to infer that it covers the whole export. Since `openssl` is already recorded and is the A53
task's own version, a consumer can close the gap itself if it is told to, so a note that
`certificates_pinned` is measured on the k3r5 root and that the A53 root's pinning follows from
`openssl` being 3.4 or newer would be enough. Either put that in a schema document alongside the
export, or rename the field to name its scope.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -306,10 +330,20 @@ tcbsgn_write_am62_manifest() {
...
+             "${outdir}/a53/tiboot3-am62x-hs-fs-verdin.bin")
```
rborn-tx: ### The manifest re-derives the a53 root instead of using `TCB_SIGNING_EXPORT_DIR_A53`

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:336`

```sh
             "${outdir}/a53/tiboot3-am62x-hs-fs-verdin.bin")
```

The file defines the root once, at line 173:

```sh
TCB_SIGNING_EXPORT_DIR_A53 ?= "${TCB_SIGNING_EXPORT_DIR}/a53"
```

and `tcbsgn_deploy_am62_export_a53` uses it, at line 208:

```sh
    local root="${DEPLOYDIR}/${TCB_SIGNING_EXPORT_DIR_A53}"
```

The new line spells `a53` out again. Because the variable is `?=`, a distro or bbappend can
change it, and then the producer writes to the new location while the manifest looks in the old
one, finds nothing, and reports `certificates_pinned: false`. The failure is silent by design,
which is what makes the coupling worth fixing rather than tolerating: there is no build error to
lead anyone back here.

The filename is duplicated in the same way. `tcbsgn_deploy_am62_export_a53` builds it in a loop
at lines 269-270 as `tiboot3-am62x-${f}-verdin.bin` over `gp hs hs-fs`, so the two spellings of
`hs-fs` have to be kept in step by hand.

Fix:

```sh
    pinned=$(tcbsgn_am62_certificates_pinned \
             "${DEPLOYDIR}/${TCB_SIGNING_EXPORT_DIR_A53}/tiboot3-am62x-hs-fs-verdin.bin")
```

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -271,14 +271,38 @@ tcbsgn_deploy_am62_export_a53() {
...
-# U-Boot produced the artifacts, so that a consumer whose binman comes from
-# another revision is detectable instead of silently producing different output.
-# The A53 side writes it alone, since everything in it is build-wide and both
-# multiconfigs build one recipe at one SRCREV.
+# Echoes whether the certificate's validity was pinned. Measured, not taken
+# from the switch that asks for it: the signing tool pins validity only when
+# its openssl is new enough, so intent and outcome can differ. Pass a k3r5
+# artifact: the A53's own outputs are FIT images, whose certificates are not
+# at offset 0. Unreadable is "false", the answer that costs a consumer least.
+tcbsgn_am62_certificates_pinned() {
```
rborn-tx, line 279: ### The file's comment on command substitution now contradicts the code

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:13`

```sh
# comes back in a variable because BitBake emits only the shell functions it
# sees called, and a call inside a command substitution is not one of them.
```

The new code calls `tcbsgn_am62_certificates_pinned` from inside a command substitution, so as
written the file now says its own new code cannot work. It is what made me check, and it is what
CodeRabbit checked as well before dismissing it, so the comment has already cost two readers
their time.

The claim is not simply wrong, which is the interesting part. I ran BitBake 2.8.1's
`bb.codeparser.ShellParser`, the scarthgap version, over four shapes:

| shell                          | `execs` found                            |
|--------------------------------|------------------------------------------|
| `pinned=$(helper "$x")`        | `helper` (plus `openssl`, `awk`)         |
| `helper`                       | `helper`                                 |
| `echo $(helper a b)`           | `helper`                                 |
| `x="prefix $(helper) suffix"`  | nothing                                  |

A bare `$(...)` is scanned. One nested inside double quotes is not, on 2.8.1. On BitBake master
the quoted form is found too, by the `if len(part) >= 3 and part[0] == '"'` branch in
`process_words`, which 2.8.1 does not have. So the comment reads like it was written after being
bitten by the quoted form, and then generalised one step too far.

Fix: narrow the comment to the case that is real, so the next reader does not have to redo this.
Something like "a call inside a double-quoted command substitution is not one of them, on the
BitBake in scarthgap".

_(Anchored here because the comment this is about, at `recipes-bsp/u-boot/u-boot-tcb-sign.inc:13`, is outside this PR's diff.)_

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -271,14 +271,38 @@ tcbsgn_deploy_am62_export_a53() {
...
+    local certfile notbefore epoch
+
+    certfile="$1"
+    [ -f "${certfile}" ] || { echo "false"; return 0; }
+
+    notbefore=$(openssl x509 -inform DER -in "${certfile}" -noout -startdate \
+                2>/dev/null | sed -ne 's/^notBefore=//p')
+    [ -n "${notbefore}" ] || { echo "false"; return 0; }
+
+    epoch=$(date -u -d "${notbefore}" +%s 2>/dev/null)
```
rborn-tx: ### The `epoch` guard does not survive `set -e` under dash

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:289`

```sh
    epoch=$(date -u -d "${notbefore}" +%s 2>/dev/null)
    if [ -n "${epoch}" -a "${epoch}" = "${SOURCE_DATE_EPOCH}" ]; then
```

The `-n "${epoch}"` test says a `date` failure should fall through to `false`. It does not.
BitBake writes `#!/bin/sh` and `set -e` into every task script (`shell_trap_code()` in
`bb/build.py`, both the bash and the non-bash arms), and `epoch=$(...)` is a simple command whose
status is the substitution's, so a failing `date` ends the shell.

Where it ends depends on which shell `/bin/sh` is, which differs between build hosts. I ran the
helper verbatim, called the way the manifest calls it, with `openssl` stubbed to emit an
unparseable `notBefore`:

- bash: the substitution's own subshell swallows it, the helper goes on to echo `false`, and the
  task continues.
- dash, which is `/bin/sh` on Debian and Ubuntu: the subshell exits 1, the outer
  `pinned=$(...)` inherits that, and `do_deploy` fails.

The other two substitutions in this change are safe by accident rather than by design, which is
worth knowing before anyone edits them. `notbefore=$(openssl ... | sed ...)` and
`opensslver=$(openssl version | awk ...)` are pipelines, so their status is `sed`'s and `awk`'s,
and both are 0 even when `openssl` is missing entirely.

I could not reach this on real input and I do not think anyone can: `openssl x509 -startdate`
prints a fixed format and GNU `date` parses it, which I checked for epoch 0, for a 2018 date and
for a 2026 one, under `TZ=America/Sao_Paulo`, `TZ=UTC` and `TZ=Asia/Tokyo`. So this is a nit
about the guard being dead code rather than about a failure anyone will hit. If the intent is
that the helper never aborts the task, `epoch=$(date ... ) || epoch=""` says so and is true on
both shells.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -306,10 +330,20 @@ tcbsgn_write_am62_manifest() {
...
     done
 
+    # The a53 root's own copy, installed earlier in this same append, so this
+    # reaches into no other multiconfig's deploy directory.
+    pinned=$(tcbsgn_am62_certificates_pinned \
+             "${outdir}/a53/tiboot3-am62x-hs-fs-verdin.bin")
+
+    # Provenance only: this task's openssl, which the k3r5 side takes from the
+    # same recipe at the same version. Consumers do not branch on it.
+    opensslver=$(openssl version | awk '{print $2}')
```
rborn-tx, line 342: ### The two new values reach the manifest unvalidated

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:340`

```sh
    opensslver=$(openssl version | awk '{print $2}')
```

and the `printf` at line 343, which takes `${SOURCE_DATE_EPOCH}` straight into a JSON number
position.

The same function bbfatals when `u-boot-version.json-${cfgtype}` is missing and again when
`ubootversion` or `ubootrelease` comes back empty. The two new values get neither treatment:

- An empty `SOURCE_DATE_EPOCH` writes `"source_date_epoch": ,` and the file stops being JSON. I
  confirmed both halves, that the normal output parses and that this one does not. It needs
  someone to set `SOURCE_DATE_EPOCH = ""` explicitly, since `bitbake.conf` assigns it with `?=`
  and falls back to `SOURCE_DATE_EPOCH_FALLBACK`, so it is remote.
- A missing `openssl` writes `"openssl": ""`. Valid JSON, no error, and `certificates_pinned`
  goes to `false` at the same time, so a build with no openssl looks exactly like a build that
  chose not to pin. This is also remote: `u-boot.inc` carries `PACKAGECONFIG ??= "openssl"` with
  `PACKAGECONFIG[openssl] = ",,openssl-native"`, so openssl-native is in `DEPENDS` by default,
  and the export only runs where FIT signing is on, which is the case that needs it anyway.
  Worth noting that `openssl` is in neither `HOSTTOOLS` nor `HOSTTOOLS_NONFATAL`, so there is no
  host fallback if that `PACKAGECONFIG` ever changes.

The helper also drops the required-argument idiom the rest of the file uses, `local src="${1?source file}"`
in `tcbsgn_export_install` and `local cfg="${1?property name in the U-Boot config}"` in
`tcbsgn_check_config`, in favour of a plain `certfile="$1"` at line 282. Called with no argument
it reports `false`.

Fix: `[ -n "${opensslver}" ]` and `[ -n "${SOURCE_DATE_EPOCH}" ]` guarded with `bbfatal`, matching
what the function already does two dozen lines above, and `local certfile="${1?certificate file}"`.

rborn-tx (commented):
Posting findings by Claude.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Correct, and it finds a real gap: "one source tree, one patch" covers the patch being applied, but it gates on `self.version_tuple() >= MIN_VALIDITY_VERSION`, answered by whichever openssl each multiconfig invokes from its own `recipe-sysroot-native`.

Declining, accepting the risk. It needs a deliberately asymmetric `openssl-native`, which nothing here produces; if it happened, binman warns in the A53 log, so the condition is visible where it arises, just not in the manifest; and the consumer has been told the field is a k3r5 measurement.

Your documentation point stands: this format has no schema anywhere, the original four included.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Agreed, fixed. The path is now `${DEPLOYDIR}/${TCB_SIGNING_EXPORT_DIR_A53}/...`, so producer and reader move together. Your argument from the silence is the right one: an override would leave the manifest reporting unpinned certificates for a build that had them, with nothing failing to lead anyone back here.

Not taking the filename half, though. `tiboot3-am62x-hs-fs-verdin.bin` is a fixed upstream artifact name rather than a local convention, so factoring one string used in two places adds indirection without removing a way to be wrong.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Agreed, fixed — and thank you for running the parser rather than only flagging the contradiction. The note now says scarthgap's parser misses a call inside a quoted expansion, which is the case that is actually real.

It was presumably written after being bitten by the quoted form and then generalised one step too far, exactly as your table shows. Narrowing it means the next reader does not have to repeat the experiment.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Confirmed and fixed. Reproduced it directly: `set -e` is written at `build.py:377` and `:381`, and a bare assignment from a failing command survives on bash but ends the shell on dash, so the `-n` test was describing behaviour the code did not have.

Now `epoch=$(date ...) || epoch=""`, which reads the same on both shells. Retested with a stubbed `openssl` emitting an unparseable date: both survive and return `false`, and real certificates still return `true`.

Taking your note about the two pipelines being safe by accident as a reason to guard the value they produce, below.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Three parts, three answers.

The empty `SOURCE_DATE_EPOCH` is already fixed — you reviewed the first commit; the branch now emits `null` there. Not `bbfatal`, because unlike a missing version file it leaves the manifest something true to say, and the consumer already handles it through `certificates_pinned`.

The empty `openssl` I have made fatal, on your argument: valid JSON, no error, and indistinguishable from a build that chose not to pin is exactly the ambiguity these fields exist to remove.

`${1?certificate file}` adopted; you are right that the file already had an idiom and this ignored it.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -306,10 +332,28 @@ tcbsgn_write_am62_manifest() {
...
+    if [ -z "${opensslver}" ]; then
+        bbfatal "\"openssl version\" produced nothing; the manifest records it" \
+                "as the provenance of the certificates the export carries."
+    fi
+
+    # Unset would interpolate as nothing and leave the field empty, which is not
+    # JSON. Report it unknown rather than shipping a manifest nobody can parse.
+    epoch="${SOURCE_DATE_EPOCH:-null}"
```
rborn-tx: ### `${SOURCE_DATE_EPOCH:-null}` moves the epoch expansion from BitBake to the shell

`recipes-bsp/u-boot/u-boot-tcb-sign.inc:350`, at head `08aa85cbe123c0bdeb059f46ea1a5a2efd309efe`

```sh
    # Unset would interpolate as nothing and leave the field empty, which is not
    # JSON. Report it unknown rather than shipping a manifest nobody can parse.
    epoch="${SOURCE_DATE_EPOCH:-null}"
```

Found while verifying the fix for finding 5, which this line is. It works, and the manifest it
produces is correct, so this is about how it works rather than about a wrong value today.

The line reads as a BitBake variable with a shell default. That is not what happens. BitBake's
variable regex in `data_smart.py:34` is

```python
__expand_var_regexp__ = re.compile(r"\${[a-zA-Z0-9\-_+./~:]+}")
```

and `:` and `-` are both in the character class, so the whole of `SOURCE_DATE_EPOCH:-null` is
taken as the variable name. No such variable exists, `var_sub` returns the match untouched, and
the text reaches the task script verbatim for the shell to expand against its environment.

I checked this against `bb.data_smart` from BitBake 2.8.1 with `SOURCE_DATE_EPOCH` set to `0`,
the value a real build here has, so the datastore genuinely had something to give:

| expression                   | BitBake returns              | reference recorded        |
|------------------------------|------------------------------|---------------------------|
| `${SOURCE_DATE_EPOCH:-null}` | `${SOURCE_DATE_EPOCH:-null}` | `SOURCE_DATE_EPOCH:-null` |
| `${SOURCE_DATE_EPOCH}`       | `0`                          | `SOURCE_DATE_EPOCH`       |

Two consequences follow, one inert and one not.

The inert one: BitBake now records a dependency on a variable that does not exist, and none on the
one the line is about. It changes nothing, because `SOURCE_DATE_EPOCH` is in
`BB_BASEHASH_IGNORE_VARS` in `bitbake.conf` and was excluded from the signature either way. It
would start to matter if that ever changed.

The one that is not: the manifest now depends on `SOURCE_DATE_EPOCH` being exported, and the code
it replaced did not. `bitbake.conf:685` says `export SOURCE_DATE_EPOCH ?=`, and the export flag
sticks to the variable rather than to the assignment, so an ordinary override keeps it. Dropping
the flag takes a deliberate `delVarFlag` or a change in oe-core, which makes this remote. But if
it happened, the manifest would report `null` while BitBake held a real epoch, which is precisely
the failure the line was written to prevent, and it would do so silently.

The wider point is that the line is correct by accident. It relies on BitBake failing a lookup on
a malformed name and leaving text behind that happens to be valid shell. A reader has no way to
tell that from the code.

Fix, if it is worth closing: keep the expansion on BitBake's side, where the comment implies it
already is.

```sh
    epoch="${SOURCE_DATE_EPOCH}"
    [ -n "${epoch}" ] || epoch="null"
```

This reads the datastore value, records the real reference, and gives the same four results for
unset, empty, `0` and a real epoch.

rborn-tx: Confirmed, and it is a real defect in the fix rather than a style point. Ran it against 2.8.1's `data_smart`:

```
${SOURCE_DATE_EPOCH:-null} -> '${SOURCE_DATE_EPOCH:-null}'  refs=['SOURCE_DATE_EPOCH:-null']
${SOURCE_DATE_EPOCH}       -> '0'                           refs=['SOURCE_DATE_EPOCH']
```

So BitBake recorded a dependency on a name that does not exist and none on the one the line is about, and the value arrived only because the variable happens to be exported.

Taken your fix verbatim. All four cases still give the right answer, and the comment no longer implies an expansion that was not happening.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -306,10 +332,29 @@ tcbsgn_write_am62_manifest() {
...
+    opensslver=$(openssl version | awk '{print $2}')
+    if [ -z "${opensslver}" ]; then
+        bbfatal "\"openssl version\" produced nothing; the manifest records it" \
+                "as the provenance of the certificates the export carries."
+    fi
+
+    # Empty would leave the field with no value at all, which is not JSON. A
+    # shell default here would read as BitBake's and be the shell's, so: two lines.
+    epoch="${SOURCE_DATE_EPOCH}"
+    [ -n "${epoch}" ] || epoch="null"
```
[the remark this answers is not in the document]
rborn-tx: Risk accepted.

jsrc27:
Looks good, just noticed one bit of strange wording in the commit message:

> and whether that can succeed turns(?) on the epoch the certificates were

Maybe "depends" was meant here?

rborn-tx:
> > and whether that can succeed turns(?) on the epoch the certificates were
> 
> Maybe "depends" was meant here?

@jsrc27 Nice catch :slightly_smiling_face:. Fixed!

MatheusRodrigues-tor (approved):

jsrc27 (approved):

Freireg (approved):

EdTheBearded (approved):

PR #607, opened 2026-09-17
Attempt to fix build issue in CI affecting verdin-am62 tdx-signed images

rborn-tx:
Five consecutive `verdin-am62` CI builds (531-535) fail in `tcb-signing-files:do_compile` with

```
tar: tcb-signing/a53/spl/u-boot-spl: file changed as we read it
```

The packing task tars its inputs out of `DEPLOY_DIR_IMAGE`, a directory it does not own. Of the eighteen tasks that publish there, **two** are ordered before the pack and **fourteen are unordered**: naming the producers of the packed files is not the same as naming the writers of the directory, and only the second is what a reader of that directory races with.

`do_rm_work` is one of the fourteen, and it removes the workdir the deploy directory's hardlinks point into. In **all five** failures a u-boot `do_rm_work` overlaps the sub-second pack, and the root whose workdir is being removed predicts the root `tar` names — seven files across the five builds, no misses.

**The stat-level mechanism is not explained, and this PR does not claim one.** Removing one name of a hard-linked inode changes `nlink` and `ctime`, neither of which GNU tar compares; reproductions on ext4 and overlayfs do not trip it. What is claimed is the correlation above and the ordering defect, which is worth closing on its own. The fix is therefore also the experiment: if the failure survives it, the correlation is disproved cheaply.

### Summary

- `tcb-signing-files.inc`: one list of packing machines, and the hold condition derived from it.
- `tcb-signing-files.bb`: derive `COMPATIBLE_MACHINE` from that same list so the two cannot drift.
- `u-boot-tcb-sign.inc`: hold the U-Boot workdir until the tarball is packed, on all three packing machines.
- Same file: exclude the k3r5 U-Boot from `rm_work`, where no task dependency can reach the packer.
- New bbappends hold `imx-atf` and `imx-boot-firmware-files`, the iMX8M's other two packed producers.
- `conf/layer.conf`: add the `freescale-layer` dynamic-layer entry those bbappends need.

Each commit carries its own reasoning; what follows is what the commit messages do not cover.

Related-to: TOR-4543

<details>

  ### Design notes

  **The gate has two halves and both are load-bearing.** `TCB_SIGNING_SUPPORT == "1"` — without it a signing-*off* `verdin-imx8mp` build pulls the packing recipe into the graph through this very edge, where its `do_compile` calls `bbfatal` for exactly that case. And membership of `TCB_SIGNING_PACK_MACHINES` — without it the **k3r5** multiconfig names a recipe not built there and the dependency never resolves. `COMPATIBLE_MACHINE` is built from that same list, so the gate cannot name a machine the recipe is not compatible with.

  **Silencing `tar` was rejected.** `--warning=no-file-changed` still exits 1, and `--ignore-failed-read` would ship a torn or short archive without a word.

  ### Migration

  None. This changes scheduling only: no task moves except `do_rm_work`, which has no sstate of its own, and the tarball is byte-identical.

  ### Tests

  On `verdin-imx8mp` and `verdin-am62`, both signed, plus two machines that must be unaffected.

  - **The edges exist.** `bitbake -g`: `imx-atf`, `imx-boot-firmware-files` and `u-boot-toradex` `do_rm_work` → `tcb-signing-files.do_compile` on the iMX8M; `u-boot-toradex-ti` likewise on the AM62.
  - **The k3r5 exclusion lands only where intended** — set for `verdin-am62-k3r5`, empty for `verdin-am62p-k3r5` and for the AM62's own A53, which uses the edge. After a forced rebuild that workdir survives full, where it previously held only `temp`.
  - **Nothing else moves.** `bitbake -S printdiff` on all four recipes: *"Attempted 0 tasks of which 0 didn't need to be rerun"*. The only uncacheable task is `u-boot-toradex-ti:do_rm_work`, the one that gained a varflag.
  - **The tarball is byte-identical** on both machines, before and after.
  - **Feature-off and off-list machines are untouched.** Unsigned `verdin-am62` keeps both gates empty, with `tcb-signing-files` absent from its task graph; signed `verdin-am62p` resolves empty in both multiconfigs and graphs cleanly — where a mis-written gate would fail with "Nothing PROVIDES".

  Not a rare race: builds 531-535 fail 5/5, the pack runs under a second, and `do_rm_work` overlaps all of it. The build machines never reproduce it, so the ordering above is the evidence, not an absence of failures.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and applying Freescale-specific build customizations when the Freescale layer is enabled.
  * Added machine-aware TCB signing package generation for supported devices.
  * Preserved required bootloader, firmware, and memory-training artifacts while signing packages are assembled.
* **Bug Fixes**
  * Prevented required signing artifacts from being removed during build cleanup.
  * Improved machine matching to avoid unintended partial or multiconfiguration matches.


rborn-tx:
Leaving as draft to solve LLM raised issues.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -7,6 +7,18 @@
 # list must name the same directory this file installs into.
 require recipes-bsp/tcb-signing-files/tcb-signing-files.inc
 
+# The export this file deploys is packed out of DEPLOY_DIR_IMAGE, so this
+# recipe's WORKDIR has to outlive the packing task; see the include above.
+do_rm_work[depends] += "${TCB_SIGNING_PACK_DEPENDS}"
```
rborn-tx, line 12: ### The diagnosed mechanism cannot produce the reported `tar` error, so the race is probably still open

Location: `recipes-bsp/u-boot/u-boot-tcb-sign.inc:10-12` (and the same reasoning in `recipes-bsp/tcb-signing-files/tcb-signing-files.inc:81-87`, both new bbappends, and all three commit messages).

```
# The export this file deploys is packed out of DEPLOY_DIR_IMAGE, so this
# recipe's WORKDIR has to outlive the packing task; see the include above.
do_rm_work[depends] += "${TCB_SIGNING_PACK_DEPENDS}"
```

The claim under test, from the PR description: "The packing task tars its inputs out of `DEPLOY_DIR_IMAGE`, where sstate publishes them as hardlinks into the producing recipe's `WORKDIR`". The first half is right and I confirmed it. `deploy.bbclass:7` sets `DEPLOYDIR = "${WORKDIR}/deploy-${PN}"`, `sstate.bbclass:331-333` runs `oe.path.copyhardlinktree(state[1], state[2])` from that directory into `DEPLOY_DIR_IMAGE`, and `oe/path.py`'s `copyhardlinktree` uses `cp -afl` whenever a test `link()` succeeds. So the two names do share an inode.

The second half is where it breaks. Removing one name of a hard-linked inode does not change the inode's size or mtime, and GNU tar does not look at anything else that changes. From `tar-1.35/src/create.c`, the only check that emits this message:

```c
	  /* Heuristically check whether the file is the same in all
	     attributes that tar cares about and can easily check.
	     ...
	     Do not check ctime where changes might be benign (e.g.,
	     another process creates a hard link to the file).  */
	  ok &= st1.st_uid == st2.st_uid;
	  ok &= st1.st_gid == st2.st_gid;
	  ok &= st1.st_mode == st2.st_mode;
	  if (! (is_dir && remove_files_option))
	    ok &= ! timespec_cmp (get_stat_mtime (&st1), get_stat_mtime (&st2));
	  if (!is_dir)
	    ok &= st1.st_size == st2.st_size;
	  if (!ok)
	    {
	      WARNOPT (WARN_FILE_CHANGED, (0, 0, _("%s: file changed as we read it"), ...
```

Reproduction, with the same tar flags the packing task uses and a window wide enough that the timing is not in doubt:

```
$ dd if=/dev/urandom of=work/big.bin bs=1M count=800
$ ln work/big.bin deploy/big.bin
before: ino=20373872 nlink=2 size=838860800 mtime=1789612469 ctime=1789612469

  tar start 1789612469.890781345
  rm -rf WORKDIR completed at 1789612474.892761261     # 5 s into a 19.5 s read
  tar exit=0
  tar end   1789612489.433737501

after:  ino=20373872 nlink=1 size=838860800 mtime=1789612469 ctime=1789612474
```

`tar --dereference --hard-dereference --sort=name --format=gnu -czf ... -C . deploy`, GNU tar 1.35. No warning, exit 0. Only `nlink` and `ctime` moved.

Two control experiments to show the harness does detect real interference:

- Replacing the directory entry with a link to a different inode mid-read (what sstate's `cp -afl` does to a pre-existing destination, since `-f` unlinks before linking) gives `tar: deploy: file changed as we read it` and exit 1. Note it names the *directory*: tar's open fd still holds the old inode, and what changed is the parent directory's mtime.
- A write to the shared inode mid-read gives nothing, because the size did not change.

So to get the message CI actually saw, naming a file rather than a directory:

```
tar: tcb-signing/a53/tiboot3-am62x-hs-fs-verdin.bin: file changed as we read it
```

something must change that file's own uid, gid, mode, mtime or size between tar's `fstatat` and its post-read `fstat`, or must swap the directory entry for a different inode inside the sub-millisecond window between tar's stat and its `open`. `rm -rf` on a sibling `WORKDIR` name does neither. A concurrent `sstate_install` into the shared `DEPLOY_DIR_IMAGE` does both: `copyhardlinktree` first extracts a `tar` directory pass with `-p --xattrs` over the existing tree, which rewrites directory modes and mtimes, then `cp -afl` unlinks and re-links every colliding file.

That points the investigation at "which task was writing into `DEPLOY_DIR_IMAGE` while `tcb-signing-files:do_compile` was reading", not at `rm_work`. On this machine `tcb-signing/` is written by two multiconfigs, and `do_compile[depends]` plus `do_compile[mcdepends]` are supposed to order both before the pack. One of those edges not covering what it looks like it covers would explain the failure; `rm_work` would not.

Concrete suggestion, in order:

1. Do not land this as the fix for TOR-4543's CI failure until the writer is identified. Re-read the failing job's task timeline for anything touching `DEPLOY_DIR_IMAGE` (any recipe's `do_deploy`, any `do_deploy_setscene`, any image task) that overlaps `tcb-signing-files:do_compile`, rather than for `do_rm_work` alone.
2. If it cannot be reproduced from logs, instrument it: run the packing task with `fatrace` or an `inotifywait -m -r` on `DEPLOY_DIR_IMAGE/tcb-signing` in the CI job, and let the next occurrence name the writer.
3. If you keep the `rm_work` ordering anyway as defence in depth, say so in the commit messages rather than presenting it as the cause. As written, the three commit messages and the PR description all assert a mechanism that is disprovable in ten lines of shell, which will mislead the next person who debugs this area.

I want to be clear about the limit of this finding: I have disproved the stated mechanism, not the existence of a race. The ordering hazard the description describes is real (nothing orders `do_rm_work` against another recipe's tasks), and holding the workdir is not a harmful thing to do. What I cannot support is that it fixes the observed failure.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
```diff
@@ -7,6 +7,18 @@
...
+
+# The k3r5 multiconfig deploys its half of the export into the same shared
+# directory, but the packing recipe is not built there, so no task dependency
+# can reach it. An mcdepends back into the packing multiconfig does not work
+# either: bitbake applies a multiconfig's whole mcdepends list to every task
+# that declares one, so it also lands on this recipe's do_compile and closes a
+# loop through do_deploy. Keeping the workdir for the build is what is left.
+RM_WORK_EXCLUDE:append:verdin-am62-k3r5 = "${@' ' + d.getVar('PN') if d.getVar('TCB_SIGNING_SUPPORT') == '1' else ''}"
```
rborn-tx, line 20: ### The k3r5 exclusion keeps the entire U-Boot workdir where one subdirectory would do

Location: `recipes-bsp/u-boot/u-boot-tcb-sign.inc:20`

```
RM_WORK_EXCLUDE:append:verdin-am62-k3r5 = "${@' ' + d.getVar('PN') if d.getVar('TCB_SIGNING_SUPPORT') == '1' else ''}"
```

`RM_WORK_EXCLUDE` is all-or-nothing per recipe. `rm_work.bbclass:43-49` exits the task outright, and `inject_rm_work` (`rm_work.bbclass:181-183`) never wires `do_rm_work` into `do_build` at all. The result is that a signed `verdin-am62` build retains the complete k3r5 U-Boot workdir, source tree and build tree included, for the whole build. `rm_work` is normally enabled precisely because the builder is short on disk, so this is the one place in the PR that costs something measurable on every build rather than only on the graph.

The only thing that has to survive is the hard-link farm, which lives in exactly one workdir entry: `DEPLOYDIR`, defined as `${WORKDIR}/deploy-${PN}` at `deploy.bbclass:7` and used as the sstate input directory at `deploy.bbclass:9`. `rm_work.bbclass` already has the per-entry knob for this. It reads `RM_WORK_EXCLUDE_ITEMS` at line 55 and skips matching top-level workdir entries at line 120:

```
    excludes='${RM_WORK_EXCLUDE_ITEMS}'
...
        elif ! echo "$excludes" | grep -q -w "$dir"; then
            "${RM_BIN}" -rf -- $dir
```

So the same hold costs a few MB instead of a few GB:

```
RM_WORK_EXCLUDE_ITEMS:append:verdin-am62-k3r5 = "${@' deploy-' + d.getVar('PN') if d.getVar('TCB_SIGNING_SUPPORT') == '1' else ''}"
```

I checked the two things that could break this. `grep -w` treats `-` as a non-word character, so `deploy-u-boot-toradex-ti` matches cleanly against an excludes string of `temp deploy-u-boot-toradex-ti`. And `RM_WORK_EXCLUDE_ITEMS = "temp"` is a plain assignment in a globally inherited class, so a recipe-level `:append` accumulates onto it regardless of parse order.

Worth considering the same substitution for the other three producers, since it would replace the cross-recipe `do_rm_work[depends]` edges entirely and remove the scope change the first commit documents, where building a producer now also builds the packing recipe. I did not verify that end to end and it is a larger rewrite, so treat it as a direction rather than a request.

If finding 1 dissolves the need for the hold altogether, this one goes away with it.

recipes-bsp/tcb-signing-files/tcb-signing-files.inc
```diff
@@ -77,3 +77,11 @@ TCB_SIGNING_INPUT_DEPENDS:verdin-am62 ?= "virtual/bootloader"
...
 TCB_SIGNING_INPUT_MCDEPENDS ?= ""
 TCB_SIGNING_INPUT_MCDEPENDS:verdin-am62 ?= "mc::k3r5:virtual/bootloader:do_deploy"
+
+# Every recipe whose deployed files are packed adds this to do_rm_work[depends]:
+# DEPLOY_DIR_IMAGE serves them as hardlinks into that recipe's WORKDIR, which
+# rm_work is otherwise free to remove while the tarball is still being packed.
+# Empty unless this machine both enables the export and packs a file list -- the
+# k3r5 multiconfig deploys into the shared directory but builds no packing
+# recipe to name, and with the export off the recipe must not enter the build.
+TCB_SIGNING_PACK_DEPENDS ?= "${@d.getVar('TCB_SIGNING_FILES_RECIPE') + ':do_compile' if d.getVar('TCB_SIGNING_SUPPORT') == '1' and d.getVar('TCB_SIGNING_FILELIST') else ''}"
```
rborn-tx: ### The gate hand-copies `COMPATIBLE_MACHINE`, and drift between them fails the build

Location: `recipes-bsp/tcb-signing-files/tcb-signing-files.inc:87`, against `recipes-bsp/tcb-signing-files/tcb-signing-files.bb:17`

```
TCB_SIGNING_PACK_DEPENDS ?= "${@d.getVar('TCB_SIGNING_FILES_RECIPE') + ':do_compile' if d.getVar('TCB_SIGNING_SUPPORT') == '1' and d.getVar('TCB_SIGNING_FILELIST') else ''}"
```

The `TCB_SIGNING_FILELIST` half of the condition is standing in for "is this machine in `COMPATIBLE_MACHINE`". It works today because the `TCB_SIGNING_FILELIST` overrides (lines 59, 60, 64) name exactly the three machines in `COMPATIBLE_MACHINE = "^(verdin-imx8mp|verdin-imx8mm|verdin-am62)$"`, which is what keeps the k3r5 multiconfig from naming a recipe it cannot build. The two lists are 25 lines and one file apart, with nothing tying them together.

Failure mode if they drift: add `TCB_SIGNING_FILELIST:<newmachine>` without extending `COMPATIBLE_MACHINE`, and every producer's `do_rm_work` names an unbuildable recipe. That is a hard `Nothing PROVIDES` at graph time, not a quiet skip, so it will be caught, but it will be caught by someone who has no reason to be looking at `do_rm_work`. The reverse drift, a machine added to `COMPATIBLE_MACHINE` with no filelist, silently drops the hold instead.

Fix: derive one from the other, or failing that, add a line at `COMPATIBLE_MACHINE` saying that `TCB_SIGNING_FILELIST` must cover exactly the machines it names, and why. The comment at lines 84-86 explains what the gate does; it does not warn the next editor that the two places are coupled.

conf/layer.conf
```diff
@@ -12,6 +12,7 @@ LAYERDEPENDS_meta-toradex-torizon = "sota virtualization-layer"
 LAYERSERIES_COMPAT_meta-toradex-torizon = "scarthgap"
 
 BBFILES_DYNAMIC += "\
+    freescale-layer:${LAYERDIR}/dynamic-layers/freescale/*/*/*.bbappend \
```
rborn-tx, line 15: ### The new dynamic-layer entry covers only `.bbappend`, unlike its siblings

Location: `conf/layer.conf:15`

```
    freescale-layer:${LAYERDIR}/dynamic-layers/freescale/*/*/*.bbappend \
    fsl-bsp-release:${LAYERDIR}/dynamic-layers/meta-imx-bsp/*/*/*.bb \
    fsl-bsp-release:${LAYERDIR}/dynamic-layers/meta-imx-bsp/*/*/*.bbappend \
```

Every other collection in the list that has both forms gets a `*.bb` line and a `*.bbappend` line. This one gets only the append. It is right for what the PR adds, and adding an unused `*.bb` glob would be speculative, so this is purely a note: a future `.bb` dropped under `dynamic-layers/freescale/` will be ignored without any diagnostic. Mentioning it in case the omission was not deliberate.

rborn-tx (commented):
Findings by Claude.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: You're right, and we couldn't trip it either: three runs on ext4 — including one at the real file sizes against a 9,000-entry workdir removal running for the whole tar — plus overlayfs with both names in the upper layer. Only `nlink` and `ctime` move. The commit messages and the PR description no longer claim that mechanism.

Five failing builds give more than the one did. It fails 5/5; the pack runs under a second; and a u-boot `do_rm_work` overlaps all of it in every case. The root whose workdir is being removed predicts the root `tar` names — `a53` alone in 531/533/534, and one file in *each* root in 532/535, where both workdirs go. Seven files, no misses. The file named is always among the largest in its root.

On your point 1: nothing else installs into `tcb-signing/`. Exactly two sstate manifests carry those paths, both `u-boot-toradex-ti.deploy`. And from the task graph, of the eighteen tasks that publish into `DEPLOY_DIR_IMAGE`, two are ordered before the pack and fourteen are unordered — that is the defect the PR now claims, and it stands on its own.

What we cannot explain is the stat change, so we do not. The fix removes the only variable that correlates, which makes it the experiment as well: if the next build fails the same way, the correlation is disproved in public and cheaply.

recipes-bsp/u-boot/u-boot-tcb-sign.inc
rborn-tx: Accepted, and verified before judging it: `sstate_installpkgdir` renames the tree back to `${WORKDIR}/deploy-${PN}` before `sstate_install` hardlinks it, on both the ran-locally and the from-sstate path, so that is the right entry. The cost is as you say and then some — 2.5 GB against 5.9 MB on our reference build, `recipe-sysroot-native` alone being 2.3 GB.

We are holding it until CI is green, for the reason your first finding gives. The mechanism is unexplained, and the five builds correlate at *workdir* granularity, not at *entry* granularity: they show that removing a recipe's workdir coincides with a file in its root being named, not that the hardlink farm inside it is the operative part. `RM_WORK_EXCLUDE_ITEMS` spares the farm and leaves the rest of the removal running, which is exactly right if the farm is the cause and no fix at all otherwise. Narrowing now spends the one experiment we have. It goes in as a follow-up once a few builds pass, where a regression is attributable to the narrowing.

The wider half we would rather not take: replacing the `do_rm_work[depends]` edges with retention trades an ordering constraint for a workaround, and the edges are the part that generalises.

recipes-bsp/tcb-signing-files/tcb-signing-files.inc
rborn-tx: Fixed, taking the first of your two options. Both now derive from `TCB_SIGNING_PACK_MACHINES` in the shared include: `COMPATIBLE_MACHINE` builds its anchored alternation from that list — the result is byte-identical to the literal it replaces — and the gate tests membership directly instead of inferring it from the file list.

The drift you describe is now the loud kind in both directions. A listed machine with no file list hits `do_compile`'s existing "nothing to pack" `bbfatal`, which names the variable and the machine, rather than a `Nothing PROVIDES` arriving from a `do_rm_work` edge the next editor has no reason to be reading. Checked against the old expression on all seven machine and switch combinations before committing.

conf/layer.conf
rborn-tx: Deliberate, and worth recording as such — thanks for asking. `dynamic-layers/freescale/` is there to append to `meta-freescale`'s recipes; a `.bb` under it would ship a recipe into another vendor's layer namespace, which this layer does not do. Append-only is also not new in that list: the `imx-frdm-release` entry below it is append-only for the same reason. If something ever does need a `.bb` there, the glob can land in the commit that needs it.

jsrc27 (approved):

Freireg:
@rborn-tx  just to make sure I understand correctly: is this fix still experimental at this point (i.e., it closes a real ordering defect but you can't yet confirm it's the actual cause ofthe tar failures) and CI is the real test? 
Also, do you have any theory on why this started specifically around build 531, or is the onset itself still unexplained?

EdTheBearded (approved):

rborn-tx:
> @rborn-tx just to make sure I understand correctly: is this fix still experimental at this point (i.e., it closes a real ordering defect but you can't yet confirm it's the actual cause ofthe tar failures) and CI is the real test? Also, do you have any theory on why this started specifically around build 531, or is the onset itself still unexplained?

@Freireg It started when the "tcb signing tarball generation" feature was introduced. The thing is that I couldn't reproduce it outside of CI. So, yes, this is still an attempt to solve the issue. It's based on statistics, more precisely an apparent correlation: all builds failed in CI and all of them showed the rm_work task running in parallel with the task responsible for generating the tarball (which per se shouldn't cause the issue).

rborn-tx:
Given the current approvals, I'll go ahead and merge this to see if the build issue is solved with the upcoming build over the weekend.

PR #609, opened 2026-09-18
Enable USB CDC-ACM and CDC/RNDIS net drivers for Luna SL1680

EdTheBearded:
Minor refactor of kernel fragments for synaptics device, to allow a better separation of core and community driven fragments.
Also enabled USB net drivers.

Related-to: TOR-4631

## Summary by CodeRabbit

* **Bug Fixes**
  * Applied important security fixes across supported devices, rather than limiting them to a single hardware variant.

* **New Features**
  * Added optional kernel support for PPP, serial networking, USB network adapters, USB serial devices, and related connectivity features on SL1680-based machines.
  * Added dedicated configuration areas for customer- and community-requested hardware support.

* **Documentation**
  * Improved organization and labeling of hardware-specific kernel configuration settings for clearer maintenance and customization.


jsrc27 (approved):
LGTM

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna_%.bbappend
```diff
@@ -3,16 +3,29 @@ require recipes-kernel/linux/linux-torizon.inc
...
     file://CVE-2026-43284-xfrm-esp-avoid-in-place-decrypt-on-shared-skb-frags.patch \
 "
 
-# Enable SDIO support in the device tree for Luna SL1680
+SL1680_KERNEL_CONFIG_FRAGMENTS = " \
+    file://platform/core.cfg \
+    file://platform/community.cfg \
+    file://sl1680/core.cfg \
+    file://sl1680/community.cfg \
+"
```
rborn-tx, line 16: ### Fragments sharing a basename clobber each other; all shared config is silently lost

`linux-syna` is kernel-yocto based (`linux-syna.inc` does `require recipes-kernel/linux/linux-yocto.inc`), so raw `.cfg` entries in `SRC_URI` are picked up by `find_sccs()` and handed to `scc`. `kernel-yocto.bbclass:243-250` adds `-I$(dirname $s)` for every fragment, `spp` sorts those include paths longest-first (`tools/spp:311`) and strips the longest match from each fragment path, which leaves the bare basename. `scc-cmds/kconf.cmd` then copies every fragment into one flat `${S}/.kernel-meta/configs/` directory, with the comment "we could compare the source and dest, and either warn or clobber, but for now, we just clobber".

The six new fragments live in three directories but share only two basenames, `core.cfg` and `community.cfg`, so only the last file of each name survives.

Reproduced against `kern-tools` SRCREV `7160ebe8b865dd6028aef278efa219433db93f7e`, which is what `poky/meta/recipes-kernel/kern-tools/kern-tools-native_git.bb` pins on scarthgap, using the four fragments from this PR in the SRC_URI order above.

For `MACHINE=luna-sl1680`:

```
$ cat out/config.queue
configs//./core.cfg # non-hardware
configs//./community.cfg # non-hardware
configs//./core.cfg # non-hardware
configs//./community.cfg # non-hardware
$ find out/configs -type f
out/configs/community.cfg
out/configs/core.cfg
$ head -2 out/configs/core.cfg
# Configurations required by Torizon OS for Luna SL1680
```

`luna-sl1680/core.cfg` overwrote `platform/core.cfg`, and `luna-sl1680/community.cfg` (three comment lines) overwrote `platform/community.cfg`. So Luna loses `CONFIG_OVERLAY_FS`, `CONFIG_FUSE_FS`, `CONFIG_BPF_SYSCALL`, `CONFIG_CGROUP_BPF`, `CONFIG_AUTOFS_FS`, `CONFIG_WLAN`, the MMC and reset-controller options, and all of the USB CDC-ACM / CDC / RNDIS / PPP options this PR exists to add.

For `MACHINE=sl1680` it is worse, because both surviving files are placeholders:

```
$ cat outC/configs/core.cfg
# Configurations required by Torizon OS specifically for the SL1680 vendor
# development board (not shared with other SL1680-based machines).
# (no entries yet)
```

Every option this layer contributes to the SL1680 kernel is gone.

Reachability: `do_kernel_configme` (`kernel-yocto.bbclass:472-478`) feeds `merge_config.sh` the paths from `config.queue`, which are exactly the flattened ones, so this is the real build path, not a theoretical one. It fails silently in both directions: `merge_config.sh` succeeds, and `do_kernel_configcheck` compares `.config` against the same flattened queue, so the options that never made it in are not in the set it audits and no "specified values did not make it into the kernel's final configuration" warning is produced.

Fix: give every fragment a unique basename. Verified that this is sufficient with the same tool:

```
$ bash scc --force -o outE:cfg,merge,meta -Iu u/platform-core.cfg u/luna-sl1680-core.cfg
$ find outE/configs -type f
outE/configs/platform-core.cfg
outE/configs/luna-sl1680-core.cfg
```

So keep the directories if you like the grouping and name the files `platform/platform-core.cfg`, `platform/platform-community.cfg`, `luna-sl1680/luna-sl1680-core.cfg` and so on, or drop the directories and use flat names such as `platform-core.cfg`, `luna-sl1680-community.cfg`. The flat version is simpler and the SRC_URI entries read the same.

Worth adding a build-time guard once the names are fixed, since nothing in the tooling will catch a future recurrence. A short `python () {}` in the bbappend that collects the `.cfg` basenames from `SRC_URI` and calls `bb.fatal` on a duplicate would do it.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna_%.bbappend
```diff
@@ -3,16 +3,29 @@ require recipes-kernel/linux/linux-torizon.inc
...
+
+# luna-sl1680 inherits the sl1680 override. Test MACHINE directly so that
+# SL1680-board-specific fragments do not also apply to Luna.
+SRC_URI:append = "${@bb.utils.contains('MACHINE', 'sl1680', '${SL1680_KERNEL_CONFIG_FRAGMENTS}', '', d)}"
```
rborn-tx, line 20: ### `sl2619` and the other Synaptics machines lose the Torizon-required kernel config

Before this PR, `file://fragment.cfg` sat in the unconditional `SRC_URI +=`, so it applied to every machine that builds `linux-syna` with this layer in the stack. After it, the platform fragments only arrive through the `contains('MACHINE', 'sl1680', ...)` guard or the `:luna-sl1680` override. Parse-tested with bitbake from poky scarthgap:

```
MACHINE=sl1680       -> base.patch, platform/core.cfg, platform/community.cfg, sl1680/core.cfg, sl1680/community.cfg
MACHINE=luna-sl1680  -> base.patch, 0001-dts.patch, platform/core.cfg, platform/community.cfg, luna-sl1680/core.cfg, luna-sl1680/community.cfg
MACHINE=sl2619       -> base.patch
```

`sl2619` gets nothing. It is not a hypothetical machine here: `conf/machine/include/sl2619.inc` is in this layer, `scripts/setup-environment:203` accepts it and writes `include conf/machine/include/${MACHINE}.inc` into `local.conf`, `syna-metadata` and `syna-usb-tool` both carry `COMPATIBLE_MACHINE = "(sl1680|sl2619|luna-sl1680)"`, and meta-synaptics sets `PREFERRED_VERSION_linux-syna:sl2619 = "6.12.11"`. A `sl2619` build would therefore come out without `CONFIG_OVERLAY_FS`, `CONFIG_FUSE_FS` or the BPF and cgroup options, which breaks OSTree and Docker on the resulting image, again with no build error. The same applies to meta-synaptics' `sl1680spi` and `sl1680usb`, which are SL1680-based but do not match the exact-`MACHINE` test.

This may well be deliberate, since `docs/README-syn.md` lists only `sl1680` and `luna-sl1680` as supported. If so, say it in the commit message, because the diff reads like a refactor and the coverage change is invisible in it. If it is not deliberate, the cheapest fix is to key the platform fragments off the `syna` override that `COMPATIBLE_MACHINE = "syna"` already relies on, or off `MACHINEOVERRIDES` rather than an exact `MACHINE` match.

While you are there: `platform/core.cfg` is headed "Configurations required by Torizon OS on all SL1680-based machines", which does not match a guard that only matches two machine names by exact string. Either the header or the guard should move.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/platform/community.cfg
```diff
@@ -0,0 +1,25 @@
+# Configurations for all SL1680-based machines, added on customer/community
+# request rather than being required by Torizon OS itself.
+
+# PPP (point-to-point protocol) support
+CONFIG_PPP=m
+CONFIG_PPPOE_HASH_BITS=4
+CONFIG_PPP_ASYNC=m
```
rborn-tx, line 7: ### `CONFIG_PPPOE_HASH_BITS=4` has no effect

In `drivers/net/ppp/Kconfig` at the pinned kernel revision (`66c9adc5`, checked directly), `PPPOE_HASH_BITS` is a promptless `int` whose value is derived from the `PPPOE_HASH_BITS_*` choice, with `default 4` as the fallback. A fragment cannot set a promptless symbol. On top of that the choice `depends on PPPOE`, and `CONFIG_PPPOE` is enabled neither by this fragment nor by `dolphin_defconfig`, so the whole choice is inactive.

The line is harmless but misleading, since it implies PPPoE is configured. Either drop it, or add `CONFIG_PPPOE=m` and set the choice symbol `CONFIG_PPPOE_HASH_BITS_4=y` if 16 buckets is actually wanted. Note 4 is the default either way, so the explicit setting buys nothing.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/platform/community.cfg
```diff
@@ -0,0 +1,25 @@
...
+
+# SLIP (serial line) support
+CONFIG_SLHC=m
```
rborn-tx, line 10: ### `CONFIG_SLHC=m` has no effect, and the comment above it names the wrong option

`SLHC` in `drivers/net/slip/Kconfig` is a promptless `tristate`, so a fragment cannot set it, and `CONFIG_PPP` already carries `select SLHC`. With `CONFIG_PPP=m` above, `SLHC` becomes `m` on its own.

The comment is also off: SLIP itself is `CONFIG_SLIP`, which is not enabled. `SLHC` is the Van Jacobson header-compression helper. Suggest deleting both lines, or, if SLIP is wanted, saying so with `CONFIG_SLIP=m`.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/platform/community.cfg
```diff
@@ -0,0 +1,25 @@
...
+
+# USB Device Class drivers
+CONFIG_USB_ACM=m
+
+# USB Serial Converter support
+CONFIG_USB_SERIAL_WWAN=m
+CONFIG_USB_SERIAL_OPTION=m
+
+# CRC functions
+CONFIG_CRC_CCITT=m
```
rborn-tx, line 25: ### `CONFIG_CRC_CCITT=m` is redundant

`CONFIG_PPP_ASYNC` carries `select CRC_CCITT`, so this is already implied. Unlike the `PPPOE_HASH_BITS` and `SLHC` lines above, this symbol does have a prompt, so the line works, it just does not change anything. Fine to keep as documentation of intent, fine to drop.

Freireg (approved):

rborn-tx (changes requested):
Posting findings by Claude.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna_%.bbappend
EdTheBearded: Good catch, reproduced it. Fixed by flattening to unique basenames (`synaptics-core.cfg`, `sl1680-core.cfg`, `luna-sl1680-core.cfg`, etc, no subdirs).

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna_%.bbappend
EdTheBearded: Not deliberate, fixed. Platform fragments now use `SRC_URI:append:syna` instead of the exact-MACHINE check, so `sl2619`/`sl1680spi`/`sl1680usb` get them too. Kept the exact-MACHINE check for the sl1680-only fragments since Luna also carries `sl1680` in its overrides.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/synaptics-community.cfg
```diff
@@ -0,0 +1,25 @@
+# Configurations for all SL1680-based machines, added on customer/community
+# request rather than being required by Torizon OS itself.
+
+# PPP (point-to-point protocol) support
+CONFIG_PPP=m
+CONFIG_PPPOE_HASH_BITS=4
+CONFIG_PPP_ASYNC=m
```
EdTheBearded: These came out of `menuconfig` for TOR-4631's requested options, then `diffconfig` to produce the fragment, not hand-picked. If you still think they should come out given they're no-ops, happy to resubmit without them.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/synaptics-community.cfg
```diff
@@ -0,0 +1,25 @@
...
+
+# SLIP (serial line) support
+CONFIG_SLHC=m
```
EdTheBearded: Same as above, came from menuconfig + diffconfig, not hand-picked. Happy to drop it if you'd still rather it went.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/synaptics-community.cfg
```diff
@@ -0,0 +1,25 @@
...
+
+# USB Device Class drivers
+CONFIG_USB_ACM=m
+
+# USB Serial Converter support
+CONFIG_USB_SERIAL_WWAN=m
+CONFIG_USB_SERIAL_OPTION=m
+
+# CRC functions
+CONFIG_CRC_CCITT=m
```
EdTheBearded: Same as above, came from menuconfig + diffconfig, not hand-picked. Happy to drop it if you'd still rather it went.

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/synaptics-community.cfg
rborn-tx: if that's the tool output, I wouldn't change it manually

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/synaptics-community.cfg
rborn-tx: ditto

dynamic-layers/meta-synaptics/recipes-kernel/linux/linux-syna/synaptics-community.cfg
rborn-tx: ditto

rborn-tx (approved):

Freireg (approved):

jsrc27 (approved):

EdTheBearded:
@MatheusRodrigues-tor can we merge this one?

rborn-tx:
Since this has been already approved by maintainers, I'll go ahead and merge it.

